PR #4, opened 2018-10-10
MODOAIPMH-3 Define RAML

epam-pavelk:
RAML was defined for the module


ramls/examples/resumption_tocken_error.sample
```diff
@@ -0,0 +1,9 @@
+<OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/"
```
craigmcnally: Typo in filename

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+            application/xml:
+              schema: oai-pmh
+              example: !include examples/no_records_error.sample
+
+    /{id}:
+      uriParameters:
+        id:
+          description: The unique identifier of the item in the repository
+          type: string
+      get:
```
craigmcnally: I think you might need a 422 here too in case some unknown metadataprefix is specified (cannotDisseminateFormat Error)

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+/oai:
+  /records:
+    displayName: Records
+    type:
+      list:
+        exampleCollection: !include examples/records.sample
+    get:
+      is: [ partionable, selective, metadataPrefix ]
+      responses:
+        404:
```
craigmcnally: I think you might need a 422 here too in case some unknown metadataprefix is specified (cannotDisseminateFormat Error)

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+          type: string
+          required: false
+      responses:
+        404:
+          description: "Item with a given ID and selective harvesting params (if any) not found"
+          body:
+            application/xml:
+              schema: oai-pmh
+              example: !include examples/no_record_error.sample
+        422:
```
craigmcnally: Add a 500 just in case

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+              schema: oai-pmh
+              example: !include examples/no_metadata_format_error.sample
+
+  /sets:
+    displayName: Sets
+    type:
+      list:
+        exampleCollection: !include examples/sets.sample
+    get:
+      is: [ partionable ]
```
craigmcnally: Responses?  200, 500 should probably do it.

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+    type:
+      list:
+        exampleCollection: !include examples/records.sample
+    get:
+      is: [ partionable, selective, metadataPrefix ]
+      responses:
+        404:
+          description: "Item with a given ID and selective harvesting params (if any) not found"
+          body:
+            application/xml:
```
piotr-kalashuk: According to [OAI-PMH Spec 3.1.2.1 Content-Type](http://www.openarchives.org/OAI/openarchivesprotocol.html#MIMETypes)
> The Content-Type returned for all OAI-PMH requests must be text/xml.

So probably it makes sense to use `text/xml` instead of  `application/xml`

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+    displayName: RepositoryInfo
+    get:
+      description: Returns information about a repository
+      responses:
+        200:
+          description: "Returns repository info"
+          body:
+            application/xml:
+              schema: oai-pmh
+              example: !include examples/repository_info.sample
```
piotr-kalashuk: It makes sense to follow note from [RMB ugrading notes](https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md):
>10. The resource type examples must not be strict (will result in invalid json content otherwise)
        CHANGE:
```yml
            example: <<exampleItem>>
```
        TO:
```yml
            example:
                strict: false
                value: <<exampleItem>>
```

So all the samples in the raml will be like 
```yml
              example:
                strict: false
                value: !include examples/repository_info.sample
```

ramls/oai-pmh.raml
```diff
@@ -0,0 +1,131 @@
...
+
+documentation:
+  - title: OAI-PMH Business Logic API
+    content: |
+      The Open Archives Initiative Protocol for Metadata Harvesting (OAI-PMH) provides
+      an application-independent interoperability framework based on metadata harvesting.
+      This module supports the OAI-PMH as a means of exposing FOLIO metadata.
+
+types:
+  oai-pmh: !include schemas/OAI-PMH.xsd
```
piotr-kalashuk: It makes sense to refer to OAI-PMH root element:
```yml
types:
  oai-pmh:
    type: !include schemas/OAI-PMH.xsd#OAI-PMH
    description: OAI-PMH root element
```

ramls/oai-pmh.raml
epam-pavelk: /records/{id} resource has 'metadataPrefix' trait which defines 'metadataPrefix' query param _as well as 422 status code_

ramls/oai-pmh.raml
epam-pavelk: /records resource:
- has 'metadataPrefix' trait which defines 'metadataPrefix' query param _as well as 422 status code_
- is of type 'list' which defines status code 500 among other things

ramls/oai-pmh.raml
epam-pavelk: /metadata_formats resource is of type 'list' which defines status code 500 among other things

ramls/oai-pmh.raml
epam-pavelk: /sets resource is of type 'list' which defines status codes 200 and 500 among other things

ramls/examples/resumption_tocken_error.sample
epam-pavelk: Good spot. 
Done.

ramls/oai-pmh.raml
epam-pavelk: This note is critical for EDGE module API.
This back-end module does not strictly follows OAI-PMH Spec anyway by exposing more RESTfull version of the API. Since there is no real difference between these two types (and sometimes it is even more preferable to use application/xml) I would leave it as is.

ramls/oai-pmh.raml
epam-pavelk: I agree.
Already done.

ramls/oai-pmh.raml
epam-pavelk: There is only one element defined in the schema (the root one) anyway. But I agree it makes sense to be more specific in RAML.

Done.

ramls/oai-pmh.raml
craigmcnally: I agree with @epam-pavelk here, but good to point out @piotr-kalashuk

PR #5, opened 2018-10-11
MODOAIPMH-5 Raml Module Builder

piotr-kalashuk:
- Processing descriptors
- Running RMB to process raml file
- Commenting some plugins which are for future use

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1,6 +1,6 @@
 {
   "id": "${artifactId}-${version}",
-  "name": "Real Time Availability Check Module",
+  "name": "OAI-PMH Repository Module",
   "provides": [
     {
       "id": "${artifactId}",
--- a/pom.xml
+++ b/pom.xml
@@ -42,15 +42,18 @@
 
   <properties>
     <module_name>mod-oai-pmh</module_name>
-    <aspectj.version>1.8.9</aspectj.version>
     <http.port>8081</http.port>
     <ramlfiles_path>${basedir}/ramls</ramlfiles_path>
     <ramlfiles_util_path>${ramlfiles_path}/raml-util</ramlfiles_util_path>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -211,7 +501,7 @@
     </plugins>
   </build>
 
-  <dependencyManagement>
+  <!--<dependencyManagement>
```
craigmcnally: Please remove or uncomment.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "${artifactId}-${version}",
-  "name": "Real Time Availability Check Module",
+  "name": "OAI-PMH Repository Module",
```
craigmcnally: Whoops!  How did that slip through?!

dcrossleyau:
In descriptors/ModuleDescriptor-template.json
the "id" in "provides" section should not use "artifactId" (multiple interfaces are possible)

In the POM, the "ramlfiles_util_path" and the "copy-resources-2" are superfluous because "raml-util" is properly located inside the "ramls" directory and so already copied by the previous "copy-resources".

dcrossleyau:
Also in the POM, the replace token thing for the "apidocs" might not be needed. See the example via "mod-rmb-template".

piotr-kalashuk:
Thanks @dcrossleyau for notes. The changes have been applied.

pom.xml
piotr-kalashuk: The `dependencyManagement` has been removed.

dcrossleyau:
I said:
> Also in the POM, the replace token thing for the "apidocs" might not be needed. See the example via "mod-rmb-template".

Actually something like that is needed. Sorry. However your POM section was not quite right either. This is something that most modules do not have operating properly.

I am working again on RMB-17 and RMB-18 and expect to have a solution soon.

PR #6, opened 2018-10-17
MODOAIPMH-35 implement verb Identify

piotr-kalashuk:
- The implementation of the Identify verb
- The rest of the verbs are implemented to return an error that the logic is not implemented yet
- Update raml so from/until parameters are defined as strings because if datetime is used, the parameters are passed to corresponding methods always null
- Modifying jaxb binding to use Instant as a date
- Adding maven-antrun-plugin to enable local html api console usage

### DIFF
--- a/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
+++ b/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
@@ -123,7 +123,7 @@ public class OaiPmhImpl implements Oai {
    */
   private String getErrorMessage() {
     String errorMsg = "Sorry, we can't process your request.";
-    String emails = System.getProperty("repository.adminEmails");
+    String emails = System.getProperty(REPOSITORY_ADMIN_EMAILS);
     if (emails != null && !emails.isEmpty()) {
       errorMsg += " Please contact administrator(s): " + emails;
     }
@@ -141,6 +141,6 @@ public class OaiPmhImpl implements Oai {
                         .withResponseDate(Instant.now().truncatedTo(ChronoUnit.SECONDS))
                         .withRequest(objectFactory.createRequestType()
                                                   .withVerb(verb)
-                                                  .withValue(System.getProperty("repository.baseURL")));
+                                                  .withValue(System.getProperty(REPOSITORY_BASE_URL)));
   }
 }
--- a/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
+++ b/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
@@ -7,8 +7,6 @@ import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import org.folio.oaipmh.ResponseHelper;
 import org.folio.rest.jaxrs.resource.Oai;
-import org.openarchives.oai._2.DeletedRecordType;
-import org.openarchives.oai._2.GranularityType;
 import org.openarchives.oai._2.OAIPMH;
 import org.openarchives.oai._2.OAIPMHerrorcodeType;
 import org.openarchives.oai._2.ObjectFactory;
@@ -20,15 +18,13 @@ import java.time.temporal.ChronoUnit;
 import java.util.Map;
 
 import static io.vertx.core.Future.succeededFuture;
-import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
 
 public class OaiPmhImpl implements Oai {
-  static final String REPOSITORY_NAME = "repository.name";
-  static final String REPOSITORY_BASE_URL = "repository.baseURL";
-  static final String REPOSITORY_ADMIN_EMAILS = "repository.adminEmails";
-  private static final String REPOSITORY_PROTOCOL_VERSION = "repository.protocolVersion";
-  static final String REPOSITORY_PROTOCOL_VERSION_2_0 = "2.0";
   private final Logger logger = LoggerFactory.getLogger("mod-oai-pmh");
+
+  private static final String REPOSITORY_BASE_URL = "repository.baseURL";
+  private static final String ERROR_MESSAGE = "Sorry, we can't process your request. Please contact administrator(s).";
+
   private ObjectFactory objectFactory = new ObjectFactory();
 
   @Override
@@ -49,7 +45,7 @@ public class OaiPmhImpl implements Oai {
       asyncResultHandler.handle(succeededFuture(GetOaiRecordsResponse.respond422WithApplicationXml(response)));
     } catch (Exception e) {
       logger.error("Unexpected error happened while processing ListRecords verb request", e);
-      asyncResultHandler.handle(succeededFuture(GetOaiRecordsResponse.respond500WithTextPlain(getErrorMessage())));
+      asyncResultHandler.handle(succeededFuture(GetOaiRecordsResponse.respond500WithTextPlain(ERROR_MESSAGE)));
     }
   }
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -0,0 +1,146 @@
...
+      String response = ResponseHelper.getInstance().writeToString(oai);
+      asyncResultHandler.handle(succeededFuture(GetOaiRepositoryInfoResponse.respond200WithApplicationXml(response)));
+    } catch (Exception e) {
+      logger.error("Error happened while processing Identify verb request", e);
+      asyncResultHandler.handle(succeededFuture(GetOaiRepositoryInfoResponse.respond500WithTextPlain(getErrorMessage())));
+    }
+  }
+
+  /**
+   * Generates user friendly error message in case if the error happens while the requiest is being processed
```
craigmcnally: typo request*

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -0,0 +1,146 @@
...
+   * @return user friendly error message in case if the error happens while the requiest is being processed
+   */
+  private String getErrorMessage() {
```
craigmcnally: Your error messages are so polite ;)

src/test/java/org/folio/oaipmh/ResponseHelperTest.java
```diff
@@ -0,0 +1,52 @@
...
+        .withErrors(new OAIPMHerrorType().withCode(OAIPMHerrorcodeType.BAD_VERB).withValue("error"));
+
+      String result = ResponseHelper.getInstance().writeToString(oaipmh);
+      assertNotNull(result);
+
+      // Unmarshal string to OAIPMH and verify that these objects equals
+      OAIPMH oaipmh1FromString = ResponseHelper.getInstance().stringToOaiPmh(result);
+      assertEquals(oaipmh, oaipmh1FromString);
+    } catch (JAXBException e) {
+      e.printStackTrace();
```
craigmcnally: Please use a Logger

src/test/java/org/folio/rest/impl/InitAPIsTest.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  @Test
+  public void init(TestContext context) {
+    InitAPIs initAPIs = new InitAPIs();
+    Vertx vertx = Vertx.vertx();
+    Future<Boolean> resultHandler = Future.future();
+
+    // Specifying incorrect path to config file
+    System.setProperty("configPath", "incorrectPath");
+    // Guaranty 'test.prop' is not available as Sys property. This property is defined in the 'config/config.properties' file
```
craigmcnally: guarantee

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -0,0 +1,146 @@
...
+        throw new IllegalStateException(String.format("One or more of the required repository configs missing: " +
+          "{repository.name: %s, repository.baseURL: %s, repository.adminEmails: %s}", repoName, baseUrl, emails));
+      }
+
+      OAIPMH oai = buildBaseResponse(VerbType.IDENTIFY)
+        .withIdentify(objectFactory.createIdentifyType()
+                                   .withRepositoryName(repoName + "_" + okapiHeaders.get(OKAPI_HEADER_TENANT))
+                                   .withBaseURL(baseUrl)
+                                   .withProtocolVersion(System.getProperty(REPOSITORY_PROTOCOL_VERSION, REPOSITORY_PROTOCOL_VERSION_2_0))
+                                   .withEarliestDatestamp(Instant.EPOCH.truncatedTo(ChronoUnit.SECONDS))
```
craigmcnally: Consider splitting this temporary logic out into a separate method, e.g. getEarliestDatestamp(), so that later we can swap out this for the real implementation which likely involves making calls to other modules.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  @Override
+  public void getOaiSets(String resumptionToken, Map<String, String> okapiHeaders,
+                         Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    asyncResultHandler.handle(succeededFuture(GetOaiSetsResponse.respond500WithTextPlain("The verb ListSets is not supported yet :(")));
+  }
+
+  @Override
+  public void getOaiRepositoryInfo(Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
```
craigmcnally: Consider splitting the implementation of each endpoint out into its own class.  Otherwise this class will get very long as we add implementations for the other endpoints.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
craigmcnally: The same can be said for most of these fields... eventually the values we're pulling from system properties might be pulled from mod-configuration, etc.

epam-pavelk: It make sense to do it, though the method signature will likely be changed later anyway, because we don't know yet what parameters will be required for the method to produce the value and declare it with no params for now.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
epam-pavelk: Fixed

src/test/java/org/folio/oaipmh/ResponseHelperTest.java
epam-pavelk: Fixed

src/test/java/org/folio/rest/impl/InitAPIsTest.java
epam-pavelk: Fixed

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
epam-pavelk: Done - the temporary logic was split into separate methods.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
epam-pavelk: I agree, it's a good point.
Done - the repository_info endpoint implementation was split into its own class.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -0,0 +1,146 @@
...
+      if (repoName == null || baseUrl == null || emails == null) {
+        throw new IllegalStateException(String.format("One or more of the required repository configs missing: " +
+          "{repository.name: %s, repository.baseURL: %s, repository.adminEmails: %s}", repoName, baseUrl, emails));
+      }
+
+      OAIPMH oai = buildBaseResponse(VerbType.IDENTIFY)
+        .withIdentify(objectFactory.createIdentifyType()
+                                   .withRepositoryName(repoName + "_" + okapiHeaders.get(OKAPI_HEADER_TENANT))
+                                   .withBaseURL(baseUrl)
+                                   .withProtocolVersion(System.getProperty(REPOSITORY_PROTOCOL_VERSION, REPOSITORY_PROTOCOL_VERSION_2_0))
```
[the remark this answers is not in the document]
epam-pavelk: Fixed - replaced with constant value.

epam-pavelk:
The overall changes are good.

PR #7, opened 2018-10-18
MODOAIPMH-11 OAI-PMH: Support for MARC21 XML

pautau-s:
- Implemented MarcJson to MarcXML mapper
- Create MapperFactory
- Add metadataPrefix enum
- Add marc4j dependency

### DIFF
--- a/src/main/java/org/folio/rest/utils/Mapper.java
+++ b/src/main/java/org/folio/rest/utils/Mapper.java
@@ -1,0 +1,9 @@
+package org.folio.rest.utils;
+
+import org.w3c.dom.Node;
+
+import java.io.IOException;
+
+public interface Mapper {
+  Node convert(String source) throws IOException;
+}
--- a/src/main/java/org/folio/rest/utils/MapperFactory.java
+++ b/src/main/java/org/folio/rest/utils/MapperFactory.java
@@ -1,0 +1,14 @@
+package org.folio.rest.utils;
+
+import org.folio.oaipmh.MetadataPrefix;
+
+public class MapperFactory {
+
+  public Mapper getMapper(MetadataPrefix metadataPrefix) {
+    if (MetadataPrefix.MARC_XML.equals(metadataPrefix)) {
+      return new MarcXmlMapper();
+    } else {
+      return  new XSLTMapper();
+    }
+  }
+}
--- a/src/test/java/org/folio/rest/utils/MapperFactoryTest.java
+++ b/src/test/java/org/folio/rest/utils/MapperFactoryTest.java
@@ -1,0 +1,36 @@
+package org.folio.rest.utils;
+
+import org.apache.log4j.Logger;
+import org.folio.oaipmh.MetadataPrefix;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.instanceOf;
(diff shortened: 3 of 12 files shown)


src/test/java/org/folio/rest/utils/MarcJsonToMarcXmlMapperTest.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final String SCHEMA_FILE = "ramls/schemas/MARC21slim.xsd";
+
+  @Test
+  public void correctJsonConvertingValidationTest() throws IOException {
+    logger.info("=== Test correct json file converting ===");
+    String input = getStringFromFile(StaticTestRecords.RESOURCES_CORRECT_JSON_MARC);
+    Node result = new MarcXmlMapper().convert(input);
+    assertTrue(validateDocumentAgainstSchema(result));
+  }
+  @Test
```
craigmcnally: nitpick... add an empty line

src/main/java/org/folio/rest/utils/MapperFactory.java
```diff
@@ -0,0 +1,14 @@
+package org.folio.rest.utils;
+
+import org.folio.oaipmh.MetadataPrefix;
+
+public class MapperFactory {
+
+  public Mapper getMapper(MetadataPrefix metadataPrefix) {
+    if (MetadataPrefix.MARC_XML.equals(metadataPrefix)) {
```
epam-pavelk: Better to compare enums using '==' operator.

src/main/java/org/folio/rest/utils/MapperFactory.java
```diff
@@ -0,0 +1,14 @@
...
+      return new MarcXmlMapper();
+    } else {
+      return  new XSLTMapper();
```
epam-pavelk: nitpick.. redundant whitespace

src/main/java/org/folio/rest/utils/Mapper.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.rest.utils;
+
+import org.w3c.dom.Node;
+
+import java.io.IOException;
+
+public interface Mapper {
```
epam-pavelk: I would propose to add javadocs at least for new interfaces/classes/enums (and its primary public methods) to give an idea of what the component is.

src/test/java/org/folio/rest/utils/MapperFactoryTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+
+public class MapperFactoryTest {
+
+  private static final Logger logger = Logger.getLogger(MarcJsonToMarcXmlMapperTest.class);
```
epam-pavelk: I think the logger should be of 'MapperFactoryTest'

src/main/java/org/folio/rest/utils/Mapper.java
pautau-s: Added javadocs.

PR #10, opened 2018-10-23
Modoaipmh 7 oai pmh marc to dublin core mapper

pautau-s:
- created mapper with xslt postprocessing
- moved classes to oiapmh package


src/main/java/org/folio/oaipmh/mappers/XSLTMapper.java
```diff
@@ -0,0 +1,70 @@
...
+   * @throws IllegalStateException if can't create Transformer from provided stylesheet.
+   */
+  public XSLTMapper(String stylesheet) {
+    try {
+      InputStream inputStream = Thread.currentThread().getContextClassLoader()
+          .getResourceAsStream(stylesheet);
+      TransformerFactory transformerFactory = TransformerFactory.newInstance();
+      transformerFactory.setURIResolver((href, base) -> new StreamSource(Thread.currentThread()
+          .getContextClassLoader().getResourceAsStream(href)));
+      transformer = transformerFactory.newTransformer(new StreamSource(inputStream));
```
epam-pavelk: Transformer is reusable, but not thread-safe object. If you want to compile stylesheet once and perform transformations using this pre-compiled instance, you need to use Templates object.

Here you can find more details:
https://xml.apache.org/xalan-j/xsltc_usage.html#api

src/main/java/org/folio/oaipmh/MetadataPrefix.java
```diff
@@ -1,8 +1,54 @@
...
+  MetadataPrefix(String name, Mapper mapper) {
+    this.name = name;
+    this.mapper = mapper;
+  }
+
+  public static MetadataPrefix fromName(String name) {
+    return CONSTANTS.get(name);
+  }
+
+  public static Set<String> getSupportedMetadataFormats() {
```
epam-pavelk: minor note... I would propose to rename the method to getAllMetadataFormats, for example.

src/main/java/org/folio/oaipmh/mappers/AbstractMapper.java
```diff
@@ -7,24 +14,12 @@
...
-  /**
-   * {@inheritDoc}
-   * @param source String representation of MarkJson source.
-   * @return the Node object that represents result MarcXML document
-   */
-  @Override
-  public Node convert(String source) {
-    try(InputStream inputStream
-          = new ByteArrayInputStream(source.getBytes(StandardCharsets.UTF_8))) {
+  protected Node templateConvert(String source) {
```
epam-pavelk: Instead of providing 'helper' templateConvert method, I would propose to provide default implementation of covert method that:
- does what templateConvert currently does
- and then call a kind of postProcess method which is empty by default, but can be overridden by children, e.g. XSLT transformer to perform XSLT transformation

src/main/java/org/folio/oaipmh/mappers/AbstractMapper.java
```diff
@@ -7,24 +14,18 @@
...
 
-import javax.xml.transform.dom.DOMResult;
-import java.io.*;
-import java.nio.charset.StandardCharsets;
 
-/**
- * Converts MarkJson format to MarcXML format.
- */
-public class MarcXmlMapper implements Mapper {
+public abstract class AbstractMapper implements Mapper {
```
craigmcnally: Your abstract class doesn't have any abstract methods...

craigmcnally: I think that if you're going to have an AbstractMapper class it should be generic; not specific to MARC.  

Consider the where we might need to create a mapper for something NOT derived from MarcXML... say the FOLIO Inventory json records for instance...  

If you think there is some shared/common code, I would keep it in this Base/abstract mapper class and extract the MARC-specific code out into the MarcXmlMapper class.  

If there isn't any shared/common code, then there's no need for the AbstractMapper.  Just have each mapper implement the Mapper interface. 

For the DC mapper, which builds upon the MarcXmlMapper, that's fine, just have it extend that class

src/main/java/org/folio/oaipmh/mappers/MarcXmlMapper.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.oaipmh.mappers;
+
+
+/**
+ * Converts MarkJson format to MarcXML format.
+ */
+public class MarcXmlMapper extends AbstractMapper {
```
craigmcnally: See my comments above.  As-is, this empty class doesn't serve any practical purpose

src/main/java/org/folio/oaipmh/mappers/MarcXmlMapper.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.oaipmh.mappers;
+
+
+/**
+ * Converts MarkJson format to MarcXML format.
```
craigmcnally: BTW, typo... "MarcJson"

src/main/java/org/folio/oaipmh/mappers/XSLTMapper.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.xml.transform.dom.DOMResult;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamSource;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+
+/**
+ * Converts MarkJson format to xml format.
```
craigmcnally: same typo as above

descriptors/ModuleDescriptor-template.json
```diff
@@ -101,7 +101,7 @@
       "description": "Get set structure of a repository"
     },
     {
-      "permissionName": "oia-pmh.all",
+      "permissionName": "oai-pmh.all",
```
epam-pavelk: There are several more such typos in this file. Please fix them as well.

src/main/java/org/folio/oaipmh/mappers/MarcXmlMapper.java
pautau-s: Fixed

src/main/java/org/folio/oaipmh/mappers/MarcXmlMapper.java
```diff
@@ -33,11 +36,20 @@ public Node convert(String source) {
...
     }
+  }
 
+  /**
+   * Empty post-processing.
+   *
+   * @param source Node DOM representation of MarcXml result.
+   * @return same Node
+   */
+  public Node postProcess(Node source) {
```
epam-pavelk: Let's get rid of this method since it is useless for this class.

pautau-s: Done

descriptors/ModuleDescriptor-template.json
pautau-s: Fixed

src/main/java/org/folio/oaipmh/mappers/XSLTMapper.java
pautau-s: Fixed

PR #11, opened 2018-10-24
MODOAIPMH-14 OAI-PMH: Implement verb "ListSets"

epam-pavelk:
- ListSets verb implementation added
- NEWS.md and README.md updated

### DIFF
--- a/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
+++ b/src/main/java/org/folio/rest/impl/OaiPmhImpl.java
@@ -8,6 +8,7 @@ import io.vertx.core.logging.LoggerFactory;
 import org.folio.oaipmh.Request;
 import org.folio.oaipmh.ResponseHelper;
 import org.folio.oaipmh.helpers.GetOaiRepositoryInfoHelper;
+import org.folio.oaipmh.helpers.GetOaiSetsHelper;
 import org.folio.oaipmh.helpers.VerbHelper;
 import org.folio.rest.jaxrs.resource.Oai;
 import org.openarchives.oai._2.OAIPMH;
@@ -23,6 +24,7 @@ import java.util.Map;
 
 import static io.vertx.core.Future.succeededFuture;
 import static org.openarchives.oai._2.VerbType.IDENTIFY;
+import static org.openarchives.oai._2.VerbType.LIST_SETS;
 
 public class OaiPmhImpl implements Oai {
   private final Logger logger = LoggerFactory.getLogger("mod-oai-pmh");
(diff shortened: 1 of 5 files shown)


src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -174,15 +184,49 @@ public void getOaiMetadataFormats(TestContext context) {
...
+          .body().asString();
 
     // Check that error message is returned
     context.assertNotNull(response);
 
+    // Unmarshal string to OAIPMH and verify required data presents
+    OAIPMH oaipmh1FromString = ResponseHelper.getInstance().stringToOaiPmh(response);
+
+    verifyBaseResponse(oaipmh1FromString, LIST_SETS);
+    assertThat(oaipmh1FromString.getListSets(), is(notNullValue()));
```
craigmcnally: Not a big deal, but consider using the `context.assert*` methods for consistency... Especially since the order of expected/actual is reversed here.

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -279,4 +319,13 @@ private RequestSpecification createBaseRequest() {
         .header(tenantHeader)
         .header(contentTypeHeaderXML);
   }
+
+  private void verifyBaseResponse(OAIPMH oaipmh1FromString, VerbType verb) {
```
craigmcnally: Good idea

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -23,6 +24,7 @@
 
 import static io.vertx.core.Future.succeededFuture;
 import static org.openarchives.oai._2.VerbType.IDENTIFY;
+import static org.openarchives.oai._2.VerbType.LIST_SETS;
 
 public class OaiPmhImpl implements Oai {
   private final Logger logger = LoggerFactory.getLogger("mod-oai-pmh");
```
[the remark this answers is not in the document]
epam-pavelk: Good catch. Fixed it.

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
epam-pavelk: I would prefer to use hamcrest matchers instead of context.assert* within this module if there are no reasons to do otherwise. I see the rest of the team also tend to use it in their tests. 
So I replaced all context based asserts with hamcrest versions.

PR #12, opened 2018-10-24
Modoaipmh 16 verb list metadata formats

khandramai:
- Business logic for ListMetadataFormats verb implemented
- Small update of MetadataPrefix provided (schema and metadata namespace added)


src/main/java/org/folio/oaipmh/helpers/GetOaiMetadataFormatsHelper.java
```diff
@@ -0,0 +1,142 @@
...
+      .withRequest(objectFactory.createRequestType()
+        .withVerb(VerbType.LIST_METADATA_FORMATS));
+  }
+
+  /**
+   * Creates {@link OAIPMH} with Error id-does-not-exist
+   * @return basic {@link OAIPMH}
+   */
+  private OAIPMH buildIdentifierNotFound() {
+    return objectFactory.createOAIPMH()
```
epam-pavelk: It's easier and less verbose to create JAXB object via constructor.
Not necessary to change it now, but in future I would suggest to stick to constructors.

src/main/java/org/folio/oaipmh/helpers/GetOaiMetadataFormatsHelper.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Creates {@link OAIPMH} with ListMetadataFormats element
+   * @return basic {@link OAIPMH}
+   */
+  private OAIPMH buildMetadataFormatTypesResponse() {
+    return objectFactory.createOAIPMH()
+      .withListMetadataFormats(getMetadataFormatTypes())
+      .withResponseDate(Instant.now().truncatedTo(ChronoUnit.SECONDS))
```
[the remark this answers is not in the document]
khandramai: Noted. Fixed. 
Thanks!

src/main/java/org/folio/oaipmh/helpers/GetOaiMetadataFormatsHelper.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Creates ListMetadataFormatsType of supported metadata formats
+   * @return supported metadata formats
+   */
+  private ListMetadataFormatsType getMetadataFormatTypes() {
+    return objectFactory.createListMetadataFormatsType()
+      .withMetadataFormats(objectFactory.createMetadataFormatType()
+        .withMetadataPrefix(MetadataPrefix.DC.getName())
```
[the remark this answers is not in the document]
khandramai: Noted. Done.
Thanks!

src/main/java/org/folio/oaipmh/helpers/GetOaiMetadataFormatsHelper.java
khandramai: Noted but without code changes now.
Thanks!

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -87,7 +92,19 @@ public void getOaiIdentifiers(String resumptionToken, String from, String until,
   @Override
   public void getOaiMetadataFormats(String identifier, Map<String, String> okapiHeaders,
                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    asyncResultHandler.handle(succeededFuture(GetOaiMetadataFormatsResponse.respond500WithTextPlain("The verb ListMetadataFormats is not supported yet :(")));
+
+    Request request = Request.builder().identifier(identifier).build();
```
epam-pavelk: Request holds okapi headers among other things. Since httpClient is only needed in helper, I think it's better to create it inside helper.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -87,7 +92,19 @@ public void getOaiIdentifiers(String resumptionToken, String from, String until,
...
+    HttpClientInterface client = getHttpClient(okapiHeaders);
+    GetOaiMetadataFormatsHelper helper = new GetOaiMetadataFormatsHelper(client, vertxContext, okapiHeaders);
+
+    helper.handle(request, vertxContext)
+      .thenAccept(oaipmh -> {
+        logger.info("Successfully retrieved ListMetadataFormats info: " + oaipmh);
+        asyncResultHandler.handle(succeededFuture(GetOaiRepositoryInfoResponse.respond200WithApplicationXml(oaipmh)));
```
epam-pavelk: I think GetOaiMetadataFormatsResponse should be used here.

src/test/java/org/folio/rest/impl/InventoryStorageMock.java
```diff
@@ -0,0 +1,78 @@
...
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpServer;
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.handler.BodyHandler;
+import org.apache.log4j.Logger;
+
+public class InventoryStorageMock {
```
epam-pavelk: Can we make it generic Okapi mock so that it can be reused/extended if we need to use other modules?

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
khandramai: Fixed. Thanks!

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -161,16 +173,91 @@ public void getOaiIdentifiers(TestContext context) {
...
   @Test
-  public void getOaiMetadataFormats(TestContext context) {
-    Async async = context.async();
-
-    RequestSpecification requestSpecification = createBaseRequest();
-    String response = test500WithErrorMessage(requestSpecification, LIST_METADATA_FORMATS_PATH);
+  public void getOaiMetadataFormats(TestContext context) throws JAXBException {
+    logger.info("=== Test Metadata Formats without identifier ===");
+
+    String responseWithoutIdentifier = RestAssured
```
epam-pavelk: there is convenience createBaseRequest method which might be used.

src/main/java/org/folio/oaipmh/helpers/GetOaiMetadataFormatsHelper.java
```diff
@@ -0,0 +1,142 @@
...
+
+  public GetOaiMetadataFormatsHelper(HttpClientInterface httpClient, Context ctx,
+    Map<String, String> okapiHeaders) {
+    this.httpClient = httpClient;
+    this.okapiHeaders = okapiHeaders;
+    this.ctx = ctx;
+  }
+
+  @Override
+  public CompletableFuture<String> handle(Request request, Context ctx) {
```
craigmcnally: It's redundant to pass `Context` into the constructor as well as the handler method... Please choose one.

craigmcnally:
Please push your commits w/ updates/fixes...

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
piotr-kalashuk: RestVerticle uses [CaseInsensitiveMap](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/CaseInsensitiveMap.html) to collect headers so once data added there all the keays are converted to lower case. However the Request builder creates copy of the headers `ImmutableMap.copyOf(okapiHeaders)`. After this the values can be only retrieved if the key is converted to lower case manually

epam-pavelk: I agree, the Request should not replace map implementation. I think it is already fixed in Piotr's PR.

src/test/java/org/folio/rest/impl/InventoryStorageMock.java
khandramai: Done. Thanks.

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
khandramai: Done. Thanks.

PR #13, opened 2018-10-25
MODOAIPMH-36 Preparation for implementation of the "ListIdentifiers" verb

piotr-kalashuk:
- The request@from and request@until should be strings to pass dates in the same format as in request
- Introducing ItemsStorageHelper interface with default InventoryStorageHelper implementation
- Refactor AbstractHelper so it has oai-identifier prefix initialized
- Refactor Request so it uses RequestType inside instead
- Adding oai-identifier.xsd


src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -1,32 +1,58 @@
...
+
+  /**
+   * Holds instance to handle items returned
+   */
+  protected static ItemsStorageHelper storageHelper;
+
+  /**
+   * Initializes data required for OAI-PMH responses
+   */
+  public static void init() {
```
epam-pavelk: I'm not sure if it's a good idea to introduce this method and make InitAPI to call it. Specifically because instances of the helpers which extend this AbstractHelper are created in static block, i.e. earlier than init() method is invoked.

One of the options is to initialize OAI_IDENTIFIER_PREFIX lazily in createRepositoryIdentifierPrefix():
  private String createRepositoryIdentifierPrefix() {
     if (OAI_IDENTIFIER_PREFIX == null) {
        try{
        // init OAI_IDENTIFIER_PREFIX
        return OAI_IDENTIFIER_PREFIX;
        } catch (MalformedURLException e) {
           throw new IllegalStateException(e);
        }
    } else {
       return OAI_IDENTIFIER_PREFIX;
    }
  }

Even though it might happen that several threads will try to initialize the prefix, it's not a problem since all of them will produce the same value.
The benefit is that when someone needs the prefix, it is guaranteed to be available.

@craigmcnally , @piotr-kalashuk 
what do you think?

craigmcnally: How about creating a static method to instantiate the helpers which is also called from InitAPI rather than doing so in a static block?

src/main/java/org/folio/oaipmh/helpers/ItemsStorageHelper.java
```diff
@@ -0,0 +1,41 @@
+package org.folio.oaipmh.helpers;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+import java.time.Instant;
+
+public interface ItemsStorageHelper {
```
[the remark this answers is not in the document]
piotr-kalashuk: The reason why I decided to name interface as 'Items' because according to [2.3 Item](https://www.openarchives.org/OAI/openarchivesprotocol.html#Item) section of the OAI-PMH spec:

> An _item_ is a constituent of a repository from which metadata about a resource can be disseminated. An item is conceptually a container that stores or dynamically generates metadata about a single resource in [multiple formats](https://www.openarchives.org/OAI/openarchivesprotocol.html#MetadataNamespaces), each of which can be harvested as [records](https://www.openarchives.org/OAI/openarchivesprotocol.html#Record) via the OAI-PMH. Each item has an [identifier](https://www.openarchives.org/OAI/openarchivesprotocol.html#UniqueIdentifier) that is unique within the scope of the repository of which it is a constituent.

**mod-inventory-storage** defines [instance-storage.raml](https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/instance-storage.raml) which describes endpoints to work with instances. However **mod-source-record-storage** defines [source-record-storage.raml](https://github.com/folio-org/mod-source-record-storage/blob/master/ramls/source-record-storage.raml) which contains `/record-storage/items` endpoint and looks like works with items.

Anyway we can rename the interface to `RecordsStorageHelper` which will have `items` and `records` related methods...

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
piotr-kalashuk: @craigmcnally we will introduce static method in next PR

PR #14, opened 2018-10-26
MODOAIPMH-36 Implement verb "ListIdentifiers"

piotr-kalashuk:
- The main logic implementation
- Validation of the parameters (MODOAIPMH-37)
- Handle default set (MODOAIPMH-39)
- Migrating all unit tests to junit5 and removing junit4 dependency


piotr-kalashuk:
The PR is not ready for merge as more changes will be done

piotr-kalashuk:
The unit tests resolved on unix system as well (on windows they were passing for some reason).
PR is ready for review. The only change left, is to add static method to initialize helpers which is dept from PR #13.

PR #17, opened 2018-10-29
MODOAIPMH-16 Implement verb "ListMetadataFormats"

piotr-kalashuk:
- Validate identifier format and return badArgument error with HTTP 422 code in the response
- Extract local storage identifier to build request to storage
- Update API init logic so oai-identifier prefix is being defined at start up
- Update unit tests and date time error validation for ListIdentifiers responses (MODOAIPMH-36, MODOAIPMH-37)


src/main/java/org/folio/oaipmh/helpers/InstancesStorageHelper.java
```diff
@@ -40,4 +40,11 @@ static InstancesStorageHelper getStorageHelper() {
    * @return endpoint
    */
   String buildItemsEndpoint(Request request) throws UnsupportedEncodingException;
+
+  /**
+   * Gets endpoint to search for an instance by identifier
+   * @param id instance identifier
+   * @return endpoint to search for an instance by identifier
+   */
+  String getInstanceEndpoint(String id);
```
piotr-kalashuk: I have to add separate method to search for instance instead of re-using `buildItemsEndpoint(...)` because we have to extract local storage identifier from passed identifier (which is in oai-identifier structure).

src/main/java/org/folio/oaipmh/helpers/InventoryStorageHelper.java
```diff
@@ -79,4 +79,9 @@ public String buildItemsEndpoint(Request request) throws UnsupportedEncodingExce
 
     return "/instance-storage/instances" + queryBuilder.build() + "&limit=" + RECORDS_LIMIT;
   }
+
+  @Override
+  public String getInstanceEndpoint(String id) {
+    return "/instance-storage/instances?query=source==MARC+and+id=" + id;
```
piotr-kalashuk: Due to the fact that we are interested only in marc fromat we have to specify this info to instance storage service. The `/instance-storage/instances/:id` endpoint does not allow to specify any query. So there were 3 ways to resolve this:
- Use `/instance-storage/instances` with query parameter (implemented approach)
- Use `/instance-storage/instances/:id` and check if the returned instance has "source" == "MARC"
- Use `/instance-storage/instances/:id/source-record/marc-json` to get marc record

In case the chosen approach is not efficient enough, please let me know.

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -41,19 +46,36 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
-      AbstractHelper.init();
-
       // Initialize ResponseWriter and check if jaxb marshaller is ready to operate
       if (!ResponseHelper.getInstance().isJaxbInitialized()) {
         throw new IllegalStateException("The jaxb marshaller failed initialization.");
       }
 
-      resultHandler.handle(Future.succeededFuture(true));
+      // Initialize data for OAI-PMH endpoint helpers
+      context.config().put(IDENTIFIER_PREFIX, createRepositoryIdentifierPrefix());
```
piotr-kalashuk: Instead of the static variable, I decided to introduce config usage via vert.x context. This will allow in future to implement approach to update configs [periodically](https://github.com/folio-org/raml-module-builder#adding-code-to-run-periodically).

src/main/java/org/folio/oaipmh/helpers/InventoryStorageHelper.java
epam-pavelk: The chosen approach seems to be the most efficient one.

src/main/java/org/folio/rest/impl/InitAPIs.java
epam-pavelk: That's good alternative approach. The only disadvantage is that you need access to the context to get property value.
Let's stick with it for now. If we see it causes inconveniences, will think about other ways then.

PR #19, opened 2018-10-29
MODOAIPMH-38 Date ranges

epam-pavelk:
Improve date range support - allow search by just lower or upper bound

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/CQLQueryBuilder.java
+++ b/src/main/java/org/folio/oaipmh/helpers/CQLQueryBuilder.java
@@ -55,11 +55,11 @@ public class CQLQueryBuilder {
     builder
       .append("(")
       .append(String.format("(metadata.updatedDate==null%s%s)",
-        isNotEmpty(from) ? String.format(" and metadata.createdDate>%s", from) : EMPTY,
-        isNotEmpty(until) ? String.format(" and metadata.createdDate<%s", until) : EMPTY))
+        isNotEmpty(from) ? String.format(" and metadata.createdDate>=%s", from) : EMPTY,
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/oaipmh/helpers/CQLQueryBuilder.java
```diff
@@ -30,15 +22,46 @@ public CQLQueryBuilder(boolean addQuestionMark) {
...
+   * @return {@link CQLQueryBuilder}
+   */
   public CQLQueryBuilder dateRange(String from, String until) {
-    Map<String, String> valueMap = ImmutableMap.of("from", from, "until", until);
-    StrSubstitutor sub = new StrSubstitutor(valueMap, "%{", "}");
-    builder.append(sub.replace(DATE_RANGE_STATEMENT));
+    builder
+      .append("(")
+      .append(String.format("(metadata.updatedDate==null%s%s)",
+        isNotEmpty(from) ? String.format(" and metadata.createdDate>%s", from) : EMPTY,
```
piotr-kalashuk: I've just checked specification in more details and see following in `2.7.1 Selective Harvesting and Datestamps` section:
>Range limits are inclusive: from specifies a bound that must be interpreted as "greater than or equal to", until specifies a bound that must be interpreted as "less than or equal to". Therefore, the from argument must be less than or equal to the until argument. Otherwise, a repository must issue a badArgument error.

Could you please update conditions and also validation?

Thanks in advance!

epam-pavelk: Well spotted. Fixed.
I also updated the validation.

PR #20, opened 2018-10-30
MODOAIPMH-12 Implement verb "ListRecords"

piotr-kalashuk:
- Initial change to return Records
- Moving constants to separate class as suggested by sonar
- Refactor ResponseHelper#writeToString(...) method so it throws IllegalStateException if marshal operation cannot be done

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/GetOaiRecordsHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/GetOaiRecordsHelper.java
@@ -10,29 +10,32 @@ import org.folio.oaipmh.ResponseHelper;
 import org.folio.rest.tools.client.Response;
 import org.folio.rest.tools.client.interfaces.HttpClientInterface;
 import org.folio.rest.tools.utils.TenantTool;
-import org.openarchives.oai._2.ListIdentifiersType;
+import org.openarchives.oai._2.ListRecordsType;
 import org.openarchives.oai._2.OAIPMH;
 import org.openarchives.oai._2.OAIPMHerrorType;
 import org.openarchives.oai._2.OAIPMHerrorcodeType;
+import org.openarchives.oai._2.RecordType;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/oaipmh/helpers/GetOaiRecordsHelper.java
```diff
@@ -0,0 +1,138 @@
...
+      // 2. Search for instances
+      httpClient.request(storageHelper.buildItemsEndpoint(request), request.getOkapiHeaders(), false)
+        // 3. Verify response and build list of identifiers
+        .thenApply(response -> buildRecords(request, response, ctx))
+        .thenApply(records -> {
+          if (records == null) {
+            return buildNoRecordsFoundOaiResponse(request);
+          } else {
+            // 4. Now we need to get marc data for each instance and update record with metadata
+            return buildSuccessResponse(buildOaiResponse(request)
```
piotr-kalashuk: Once the changes for [MODOAIPMH-17](https://issues.folio.org/browse/MODOAIPMH-17) are available, the logic here will be updated to make async calls to get metadata

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -675,13 +540,28 @@ void getOaiRepositoryInfo(VertxTestContext testContext) throws JAXBException {
     testContext.completeNow();
   }
 
-  private RequestSpecification createBaseRequest() {
+  @ParameterizedTest
+  @ValueSource(strings = { REPOSITORY_NAME, REPOSITORY_BASE_URL, REPOSITORY_ADMIN_EMAILS })
+  void getOaiRepositoryInfoMissingConfigs(String prop) {
```
epam-pavelk: Good idea - parameterized test seems to fit well for such case.

PR #21, opened 2018-11-01
Modoaipmh 17 oai pmh implement verb get record, MODOAIPMH-12 Implement verb ListRecords

pautau-s:
- Created abstract GetRecords helper with common for ListRecords and GetRecord logic
- Implemented GetRecord and ListRecords helpers
- Added unit support for marc record storage to OkapiMockServer
- Add 400 bad request for getRecord in raml file
- Removed unnecessary dependencies and permissions from ModuleDescriptor


src/main/java/org/folio/oaipmh/helpers/GetOaiRecordHelper.java
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      errors.add(new OAIPMHerrorType().withCode(BAD_ARGUMENT).withValue
+        (RECORD_METADATA_PREFIX_PARAM_ERROR));
+    }
+    return errors;
+  }
+
+  @Override
+  protected void addRecordsToOaiResponce(OAIPMH oaipmh, Collection<RecordType> records) {
+    oaipmh.withGetRecord(new GetRecordType().withRecord(records.iterator().next()));
```
piotr-kalashuk: There might be potential case when identifier exists but there is no mrc-json record. In this case `records` will be empty and `NoSuchElementException` will be thrown

src/main/java/org/folio/oaipmh/Constants.java
```diff
@@ -2,6 +2,10 @@
 
 public final class Constants {
 
+  private Constants() {
+    throw new IllegalStateException("This class holds constants only");
```
craigmcnally: I never thought to do this, always just used an empty constructor, but I like it.

PR #22, opened 2018-11-02
Modoaipmh 10 oai pmh incomplete response & resumable tokens

epam-pavelk:
Flow Control (resumptionToken) support  was added


src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
```diff
@@ -98,34 +123,98 @@ private OAIPMH buildNoRecordsFoundOaiResponse(Request request) {
...
-      return identifiers;
+        .map(instance -> populateHeader(identifierPrefix, (JsonObject) instance))
+        .forEach(identifiers::withHeaders);
+
+      String resumptionToken = buildResumptionToken(request, instances, totalRecords);
+
+      if (resumptionToken != null) {
+        identifiers.withResumptionToken(new ResumptionTokenType()
+          .withValue(resumptionToken)
+          .withCompleteListSize(BigInteger.valueOf(totalRecords))
```
craigmcnally: Do we really need BigInteger here, or would long suffice?

PR #23, opened 2018-11-05
Modoaipmh 54 improve marshalling of oai pmh responses to avoid redundant namespaces

pautau-s:
- Changed mappers return type to byte[]
- Added unmarshling of oairecord metadata to AbstractGetRecordHelper
- Moved the creation of marshaller and unmarshaller from the constructor to the methods
- Added missing permission to module descriptor

### DIFF
--- a/src/main/java/org/folio/oaipmh/mappers/Mapper.java
+++ b/src/main/java/org/folio/oaipmh/mappers/Mapper.java
@@ -1,16 +1,14 @@
 package org.folio.oaipmh.mappers;
 
-import org.w3c.dom.Node;
-
 /**
  * Converts MarcJson format to XML format.
  */
 public interface Mapper {
   /**
-   * Converts json string to DOM Node representation of XML document.
+   * Converts json string to byte[] representation of XML document.
    *
    * @param source String representation of MarcJson source.
-   * @return the Node object that represents result XML document.
+   * @return byte[] that represents result XML document.
    */
-  Node convert(String source);
+  byte[] convert(String source);
 }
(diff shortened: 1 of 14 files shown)


src/main/java/org/folio/oaipmh/mappers/Mapper.java
```diff
@@ -1,16 +1,14 @@
 package org.folio.oaipmh.mappers;
 
-import org.w3c.dom.Node;
-
 /**
  * Converts MarcJson format to XML format.
```
craigmcnally: fix this description...

craigmcnally: Never mind, I guess it's generic enough.

src/main/java/org/folio/oaipmh/mappers/MarcXmlMapper.java
```diff
@@ -24,33 +22,26 @@
...
-         * which contradicts to the MARC21slim.xsd schema. So replacing unexpected char by space
-         */
-        for (DataField data : record.getDataFields()) {
-          if (data.getIndicator1() == '\\') {
-            data.setIndicator1(' ');
-          }
-          if (data.getIndicator2() == '\\') {
-            data.setIndicator2(' ');
-          }
+      for (DataField data : record.getDataFields()) {
```
craigmcnally: The comments here were actually somewhat helpful... you should consider keeping them.  

Also, fixing up the two indicator fields isn't good enough here.  The backslash is used in place of a space through the MARC record.  Converting these to slashes only in some places of the record introduces inconsistencies.  Doing a global search/replace on the entire record would produce a much cleaner result.

pautau-s: Done

PR #27, opened 2018-11-09
Modoaipmh 13 oai pmh backend configuration

pautau-s:
- Implemented configuration helper
- Added test for new helper
- Added permission to access mod-configuration


src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
```diff
@@ -0,0 +1,84 @@
...
+      future.complete(null);
+    }
+    return future;
+  }
+
+  public static String getProperty(String name,Context ctx) {
+    return ctx.config().getString(name, System.getProperty(name));
+  }
+
+  public static RepositoryConfigurationHelper getInstance() {
```
craigmcnally: Is there any reason to not use a standard constructor?  A static `getInstance()` method suggests this is a singleton, which it isn't.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -53,154 +54,169 @@ public static void init(Handler<AsyncResult<Boolean>> resultHandler) {
...
+              return null;
+            });
+        } catch (Exception e) {
           asyncResultHandler.handle(succeededFuture(GetOaiRecordsByIdResponse.respond500WithTextPlain(GENERIC_ERROR_MESSAGE)));
-          return null;
-        });
-    } catch (Exception e) {
-      asyncResultHandler.handle(succeededFuture(GetOaiRecordsByIdResponse.respond500WithTextPlain(GENERIC_ERROR_MESSAGE)));
-    }
+        }
```
craigmcnally: I think you might want an `.exceptionally(...)` here...

pautau-s: added

PR #33, opened 2018-11-15
MODOAIPMH-61 Add oai-identifier description in Identify response

piotr-kalashuk:
- Implementing oai-identifier description
- Extend RepositoryConfigurationHelper to load config per tenant (initial commit)
- Introducing config for test tenant to guarantee there is no any overlap with other test configs
- Add additional logging to InitAPIs to see what configs were loaded from file and which are from VM options


src/test/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelperTest.java
```diff
@@ -45,22 +45,17 @@
...
       .put("http.port", okapiPort);
     DeploymentOptions opt = new DeploymentOptions().setConfig(conf);
 
-    vertx.deployVerticle(RestVerticle.class.getName(), opt, testContext.succeeding(id ->
-      OaiPmhImpl.init(testContext.succeeding(success -> {
-        logger.info("mod-oai-pmh Test: setup done.");
-        okapiMockServer.start(testContext);
-      }))
-    ));
+    vertx.runOnContext(event -> testContext.verify(() -> okapiMockServer.start(testContext)));
```
craigmcnally: Is the verticle deployed automatically here?

src/test/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelperTest.java
```diff
@@ -45,22 +45,17 @@
   @BeforeAll
   static void setUpOnce(Vertx vertx, VertxTestContext testContext) {
     okapiHeaders.put(OKAPI_TOKEN, "eyJhbGciOiJIUzI1NiJ9");
-    okapiHeaders.put(OKAPI_URL, "http://localhost:" + mockPort);
     OkapiMockServer okapiMockServer = new OkapiMockServer(vertx, mockPort);
     JsonObject conf = new JsonObject()
```
craigmcnally: are conf and opt still needed here?

src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
```diff
@@ -70,7 +70,15 @@
             .forEach(o ->
               config.put(((JsonObject) o).getString("code"),
                 ((JsonObject) o).getString("value")));
+          // this should be removed once proper support of the config per tenant is added
```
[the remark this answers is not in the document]
piotr-kalashuk: It is correct, however the approach is to store configs once they are retrieved to vert.x context so it can be accessible by main business logic. As far as I can see the context is shared between different calls to module that is why the configs are shared between all the requests

src/test/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelperTest.java
piotr-kalashuk: Actually the RestVerticle is not needed here but only MockServer. The RepositoryConfigurationHelper only sends requests to mod-configuration and returns result in Future. So I removed it to reduce unit test time run...

src/test/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelperTest.java
piotr-kalashuk: You are right, it is not required anymore. Thanks

piotr-kalashuk: Removed

src/test/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelperTest.java
craigmcnally: ok, makes sense

src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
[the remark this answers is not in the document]
craigmcnally: got it.  I somehow missed that we were storing the configuration on the vertx context.  Does this have the side effect of using the "last known" configuration if mod-configuration is temporarily unreachable?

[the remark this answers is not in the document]
piotr-kalashuk: Yes, once the configs are stored, they are available all the time in the context. The logic here is to use `merge in` approach which overrides old values by new ones. In case there are no configs retrieved from mod-configs or some config was deleted, the old configs (by key) will be still stored in context. To be honest I am not sure if someone is really going to remove completely configs to rely on system default values...

PR #34, opened 2018-11-19
MODOAIPMH-62-OAI-PMH-Backend-Configuration-store-configs-per-tenant

pautau-s:
- RepositoryConfigurationHelper for each tenant keeps its configuration in VertxContext

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
@@ -1,6 +1,7 @@
 package org.folio.oaipmh.helpers;
 
 import io.vertx.core.Context;
+import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationHelper.java
```diff
@@ -89,19 +85,14 @@
...
   /**
    * Gets value of the config either from shared config or from System properties as a fallback.
    * @param tenant tenant
    * @param name config key
-   * @param ctx Vert.x context
    * @return value of the config either from shared config if present. Or from System properties as fallback.
    */
-  public static String getProperty(String tenant, String name, Context ctx) {
-    JsonObject configs = ctx.config().getJsonObject(tenant);
+  public static String getProperty(String tenant, String name) {
```
epam-pavelk: The class seems to be a bit inconsistent - this method is declared as static, but another method is regular one, though both are utility methods.
I would propose the following changes:
1. Rename the class to RepositoryConfigurationUtil
2. Make both methods static
3. Rename the first method from getConfiguration to loadConfiguration since it does not return config

PR #35, opened 2018-11-21
MODOAIPMH-57 Migrate to SRS

piotr-kalashuk:
- Add `source-record-storage` dependency
- Introduce `SourceRecordStorageHelper` to work with SRS module
- Refactor storage helpers: move to separate package, introduce `AbstractStorageHelper` with shared logic for Inventory and SRS storages
- Formatting the from/until date to include `000` milliseconds while building query to storage
- ListRecords logic updated to try to build metadata records from the initial response. In case no metadata build after initial response, try to get record individually from storage
- According to OAI-PMH specification the repository must support date granularity. So handling requests if date specified in from/until parameter


src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
```diff
@@ -108,37 +110,44 @@
...
     requiresSuccessStorageResponse(instancesResponse);
 
-    JsonArray instances = storageHelper.getItems(instancesResponse.getBody());
-    Integer totalRecords = storageHelper.getTotalRecords(instancesResponse.getBody());
+    JsonObject body = instancesResponse.getBody();
+    JsonArray instances = storageHelper.getItems(body);
+    Integer totalRecords = storageHelper.getTotalRecords(body);
+
+    if (logger.isDebugEnabled()) {
+      logger.debug(String.format("%d entries retrieved out of %d", instances.size(), totalRecords));
```
epam-pavelk: It's better to use version of debug method that takes parameters instead of building log string in advance. In this case, no check for enabled debug is required.

src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
```diff
@@ -108,37 +110,44 @@
...
-                                                       .collect(Collectors.toList()));
-                if (resumptionToken != null) {
-                  addResumptionTokenToOaiResponse(oaipmh, resumptionToken, request, totalRecords);
-                }
-                return oaipmh;
-              })
-              .thenCompose(oaipmh -> supplyBlockingAsync(ctx, () -> buildResponse(oaipmh)));
-    }
+    /*
+    * According to OAI-PMH guidelines: it is recommended that the responseDate reflect the time of the repository's clock at the start
```
epam-pavelk: Good catch

src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
piotr-kalashuk: Done

PR #41, opened 2019-02-14
MODOAIPMH-66 Upgrade to source-record-storage 2.0 interface

piotr-kalashuk:
Updating SRS interface version to v2 and updating required permissions


piotr-kalashuk:
Just merging module descriptor for now to allow integration environment to be deployed. Actual functionality updates will be done as separate PR

PR #44, opened 2019-07-09
MODOAIPMH-72 Update inventory interface version

piotr-kalashuk:
## Purpose
[MODOAIPMH-72](https://issues.folio.org/browse/MODOAIPMH-72) Update inventory interface version
## Approach
- Adding `7.0` dependency along with `6.0` for `instance-storage` interface (it is safe because the logic does not work with notes)
- Just in case removing `notes` from mock files


piotr-kalashuk:
> I noticed that some mock files still have empty notes arrays... was this intentional?

Taking into account that notes are going to be still an array (but element type is changing from string to object) I decided that having empty arrays is just fine (and it reduces number of files in the PR so it is easier to review :) )

PR #46, opened 2019-07-15
MODOAIPMH-74 - Release 1.0.3

KVupp:
## Purpose
https://issues.folio.org/browse/MODOAIPMH-74 - 1.1.0 Release

## Approach
The dependencies are already in required versions
Updating NEWS


NEWS.md
```diff
@@ -1,3 +1,14 @@
+## 1.0.3 (Released)
```
javalkar-varun: Shouldn't there be an ` ## Unreleased` before `Released` section for eg: `## 1.1.0 - Unreleased` ?

pautau-s: We discussed with Piotr and decided that since we have started to support the new version of the instance-storage, the logic has changed slightly, as well as the configuration-client and raml-module-builder versions have changed, we should release it under version 1.1.0.

PR #49, opened 2019-10-31
MODOAIPMH-76  Release 1.1.1

javalkar-varun:
**Purpose:**
https://issues.folio.org/browse/MODOAIPMH-76 - release 1.1.1

**Approach:**
Update NEWS.md
Update SNAPSHOT version in pom.xml

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,14 @@
-## 1.1.1 (UnReleased)
+## 1.1.2 (UnReleased)
+
+## 1.1.1 (Released)
+Handle breaking changes in SRS and as a result update interface dependencies to avoid breaking mod-oai-pmh.
+Add Launch descriptor settings
+
+[Full Changelog](https://github.com/folio-org/mod-oai-pmh/compare/v1.1.0...v1.1.1)
+
+### Stories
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-oai-pmh</artifactId>
-  <version>1.2.0-SNAPSHOT</version>
```
khandramai: No need to make this now. Version changing will happen during release plugin execution.

NEWS.md
```diff
@@ -1,4 +1,14 @@
-## 1.1.1 (UnReleased)
+## 1.1.2 (UnReleased)
+
+## 1.1.1 (Released)
```
khandramai: Please, be careful with versioning. You can use semantic versioning guide https://semver.org/spec/v2.0.0.html. So here I don't see reasons for patch number increasing. It seems minor version should be increased, i.e. 1.2.0.

khandramai:
JIC, release procedure described here https://wiki.folio.org/display/FOLIJET/Acquisitions+-+Release+Procedures.

javalkar-varun:
Thanks for your comments @khandramai. Since we are not releasing this sprint, I'll close this PR and we can revisit when the time comes.

PR #57, opened 2020-03-30
MODOAIPMH-69: Change record identifier to use the UUID of the instanc…

IlliaDaliek:
### Purpose
To use only inventory instance id within record identifier of oai-pmh record(s) response.


src/main/java/org/folio/oaipmh/helpers/storage/SourceRecordStorageHelper.java
```diff
@@ -28,10 +35,38 @@ public String getRecordId(JsonObject entry) {
...
+    return parsedRecordFields
+      .flatMap(this::getNineNineNineField)
+      .map(field -> field.getJsonArray(SUBFIELDS))
+      .map(subfields -> subfields.getJsonObject(INSTANCE_ID_POSITION))
+      .map(instanceId -> instanceId.getString(INSTANCE_ID))
+      //if 999 field is missed then it means that record hasn't parsedRecord metadata and will be filtered from result response later
+      .orElse("");
+  }
+
+  private Optional<JsonObject> getNineNineNineField(JsonArray jsonArray){
```
mrdpopov: Can you name this method have a more meaningful name?

src/test/java/org/folio/oaipmh/helpers/StorageHelperTest.java
```diff
@@ -93,6 +98,18 @@ void getItemId(String storageType) {
...
   }
 
+  @Test
+  void shouldReturnLinkedToRecordInstanceId_whenGetRecordIdAndStorageIsSRS(){
+    JsonObject item = getJsonObjectFromFile(getDirPath(SOURCE_RECORD_STORAGE)+"/instance.json");
+    assertEquals(INSTANCE_ID, getStorageHelper(SOURCE_RECORD_STORAGE).getIdentifierId(item));
+  }
+
+  @Test
+  void shouldReturnEmptyString_whenGetRecordIdAndStorageIsSRSAndRecordHasNot999FieldWithLinkedInstanceId(){
```
mrdpopov: Are you sure that this is the correct behavior? The response from oai-pmh will have an empty id then, won't it?

src/main/java/org/folio/oaipmh/helpers/storage/SourceRecordStorageHelper.java
IlliaDaliek: Yes, sure

src/test/java/org/folio/oaipmh/helpers/StorageHelperTest.java
IlliaDaliek: This is the righ behaviour. In case if record identifier are empty it means thar record doesn't have the metadata ("parsedRecord" part of json body) and such records will be filtered from response. So, there never won't be records with empty identifier.

PR #58, opened 2020-03-30
MODOAIPMH-68: Suppress instance records from discovery

IlliaDaliek:
### PURPOSE
All SRS records/Inventory instances with suppress discovery flag = true should be ignored by OAI-PMH and should be excluded from aoi-pmh records/identifiers response.


mrdpopov:
There are changes in queries to SRS and inventory, but there are no tests for the inventory. Could one be added please?

PR #59, opened 2020-04-01
fix getRecord & getIdentifiers to use insatnce id instead of record i…

IlliaDaliek:
### PURPOSE
Fix the getRecord and getIdentifiers verbs to use insatnce id wihtin response identifier insetad of record id


mrdpopov:
Are there any tests that show that the functionality is now working correctly for all the verbs?

PR #60, opened 2020-04-09
Modoaipmh 100

IlliaDaliek:
### PURPOSE
Before enabling module for tenant performs loading of oai-pmh module configurations(properties) from mod-configuration and sets them to the system properties. In case of their absence within mod-configuration performs posting of default configurations to mod-configuration and by the end adjsuts them as default values to the system properties.  Default configurations are stored within mod-oai-pmh resource folder as json files.


mrdpopov:
It looks like we are not using the existing config.properties anymore. Please update the code. Please update news.md and add documentation to readme.md.

KITSViktorSkibnev:
What about code smells ? We can fix it ?

IlliaDaliek:
> What about code smells ? We can fix it ?
Thank you for note, already fixed

PR #67, opened 2020-05-05
MODOAIPMH-124: Add the suppressDiscovery flag into OAI-PMH feed for o…

IlliaDaliek:
### PURPOSE
Put additional field (has name "rights") with information about record discovery suppression status when repository.suppressedRecordsProcessing property has 'false' value. In opposite case just skip suppressed records and do not put any information about such record status.
pr_screen


src/main/java/org/folio/oaipmh/helpers/records/impl/DcRecordHelper.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public void updateRecordCollectionWithSuppressDiscoveryData(Collection<RecordType> records) {
+    records.forEach(record -> {
+      boolean isSuppressedFromDiscovery = record.isSuppressDiscovery();
+      String suppressDiscoveryValue = isSuppressedFromDiscovery ? "discovery suppressed" : "discovery not suppressed";
+      Dc dcRecord = (Dc) record.getMetadata().getAny();
+      List<JAXBElement<ElementType>> list = dcRecord.getTitlesAndCreatorsAndSubjects();
+      QName name = new QName(NAMESPACE_URL, "rights");
+      ElementType elementType = new ElementType().withValue(suppressDiscoveryValue);
```
craigmcnally, line 26: Instead of doing this here, does it make sense to add update the marc -> dc xslt?  See https://github.com/folio-org/mod-oai-pmh/blob/fb1a772a45ec28d8b372c6b4f8fa57866751c988/src/main/resources/xslt/MARC21slim2OAIDC.xsl#L191

craigmcnally: so I think it's being stored in 999$t as 1/0...  so something like:

```
        <xsl:for-each select="marc:datafield[@tag=999]">
            <dc:rights>
                <xsl:value-of select="marc:subfield[@code='t']"/>
            </dc:rights>
        </xsl:for-each>
```

craigmcnally: I see now that you don't want to use the 1/0 directly in the DC, so you'll probably need to use something like a choose...

```
        <xsl:for-each select="marc:datafield[@tag=999]">
            <dc:rights>
                <xsl:variable name="suppress" select="marc:subfield[@code='t']" />
                <xsl:choose>
                    <xsl:when test="$suppress > 0">
                        suppressed from discovery
                    </xsl:when>
                    <xsl:otherwise>
                        not suppressed from discovery
                    </xsl:otherwise>
                </xsl:choose>            
            </dc:rights>
        </xsl:for-each>
```

IlliaDaliek: @craigmcnally ,thank you for advice, but i prefer to add such field via the code by the reason that such field is optinal, i mean it depeneds on "suppress discovery processing setting". And if i'll do like you suggested then instead of adding the field it will be required to filter(suppress) him from the response each time.

craigmcnally: I was thinking that the marc would be conditionally updated (based on suppress discovery processing setting) with the suppress discovery field 999$t, and then the style sheet is applied for oai_dc.  

The benefit of this approach is that the dc processing is still very straightforward - take the marc and run it through an xslt, instead of having this combination of java and xslt.

src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
```diff
@@ -193,78 +193,82 @@
...
-  private void updateGeneralInfoDataFieldWithSuppressDiscoverySubfield(final Predicate<DataFieldType> generalInfoDataFieldPredicate, final String suppressDiscoveryValue, final List<DataFieldType> datafields) {
-    datafields.stream()
-      .filter(generalInfoDataFieldPredicate)
-      .findFirst()
-      .ifPresent(dataFieldType -> {
-        List<SubfieldatafieldType> subfields = dataFieldType.getSubfields();
-        subfields.add(new SubfieldatafieldType().withCode(INSTANCE_SUPPRESS_FROM_DISCOVERY_SUBFIELD_CODE)
-          .withValue(suppressDiscoveryValue));
-      });
+  @NotNull
```
craigmcnally, line 219: I'm not familiar with this, but a quick search suggests that this is specific to intellij IDEA?  Is that actually the case?  Does this have any impact at runtime?

IlliaDaliek: Hello, no it doesn't, i have deleted it

PR #68, opened 2020-05-14
MODOAIPMH-112 Fix issue with the chain of records getting with resumptionToken

KITSViktorSkibnev:
### PURPOSE
Add to the request to source-record-storage which is created in mod-oai-pmh until parameter with the current date.
### LINK
[https://issues.folio.org/browse/MODOAIPMH-112](url)

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/storage/AbstractStorageHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/storage/AbstractStorageHelper.java
@@ -1,6 +1,7 @@
 package org.folio.oaipmh.helpers.storage;
 
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.folio.oaipmh.Constants.ISO_UTC_DATE_ONLY;
 import static org.folio.oaipmh.Constants.OKAPI_TENANT;
 import static org.folio.oaipmh.Constants.REPOSITORY_MAX_RECORDS_PER_RESPONSE;
 import static org.folio.oaipmh.Constants.REPOSITORY_SUPPRESSED_RECORDS_PROCESSING;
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/oaipmh/helpers/StorageHelperTest.java
```diff
@@ -53,6 +58,8 @@
 
   private static final String INSTANCE_ID = "00000000-0000-4000-a000-000000000000";
 
+  private final DateTimeFormatter DateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
IlliaDaliek, line 61: camel case?

src/main/java/org/folio/oaipmh/helpers/storage/AbstractStorageHelper.java
```diff
@@ -60,6 +62,10 @@ protected String buildSearchQuery(Request request) throws UnsupportedEncodingExc
       queryBuilder
         .and()
         .addStrictCriteria(getIdentifierName(), request.getStorageIdentifier());
+    } else if (request.getFrom() == null && request.getUntil() == null){
+      queryBuilder
+        .and()
+        .dateRange(null, LocalDateTime.now().format(ISO_UTC_DATE_ONLY));
```
mreno-EBSCO, line 68: What date are you trying to represent here? The date on the server running `mod-oai-pmh` (which is what `LocalDateTime.now()` will return), the date local to the tenant that is being harvested, the date local to the client harvesting the data, or something else? I see the formatter constant name mentions UTC, but the formatter just includes date components, which would still use the local time zone and not UTC. If it should be UTC, we could do something like:
```suggestion
        .dateRange(null, LocalDateTime.now(ZoneOffset.UTC).format(ISO_UTC_DATE_ONLY));
```

I'm worried about confusing results when running in different timezones. Maybe this is not a problem here, but local time zone usage tends to be a problem for other modules where they typically meant to use UTC.

I see `LocalDateTime.now()` in the unit tests as well. Has this been tested in multiple time zones? We've had cases where local dates were used and unit tests depending on the time of day and the time zone where the build is executed. Again, this may not be a problem here, it is just something that I have come across in other modules that do date/time based operations and wanted to call attention to it.

PR #70, opened 2020-05-25
MODOAIPMH-131: Move request validation and configuration settings logic from edge to mod-oai-pmh

IlliaDaliek:
### PURPOSE
Move request validation from edge-oai-pmh to mod-oai-pmh. As well implement both repositoryRequestProcessing and enableOaiService settings. Perform some refactoring -> move responses building from handlers to a new responsible for it class.


src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -53,6 +82,44 @@ public static void init(Handler<AsyncResult<Boolean>> resultHandler) {
     resultHandler.handle(succeededFuture(true));
   }
 
+  @Override
+  public void getOaiVerbs(String verbName, String identifier, String resumptionToken, String from, String until, String set, String metadataPrefix, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
mrdpopov, line 86: I think we need to make up a more appropriate name for this method, as it is very confusing.

IlliaDaliek: > I think we need to make up a more appropriate name for this method, as it is very confusing.

Okay, what you can suggest for example?

mrdpopov:
Let's also remove the old rest api leaving only one entry point in mod-oai-pmh: GET oai/records

craigmcnally:
> Let's also remove the old rest api leaving only one entry point in mod-oai-pmh: GET oai/records

Maybe this should be split into a separate PR... This one already spans 30 files and 1500+ line changes.  Making it any larger will make it difficult to review without a very significant time investment.

It's unfortunate that the decision to not grant the institutional user permission to read configuration from the edge module has snowballed into all of these changes.  Just curious - @mrdpopov would you consider this refactoring of the mod-oai-pmh API a "must have" or a "nice to have"?

mrdpopov:
> Just curious - @mrdpopov would you consider this refactoring of the mod-oai-pmh API a "must have" or a "nice to have"?

My concern was with the name of the new endpoint, which was "getOaiVerbs". 
Agree that that's a lot of work involved in this refactoring, and a big chunk of line changes relate to imports. As far as I know, the removing of the old rest api is almost done.

craigmcnally:
AFAICT endpoints were removed from the RAML but not from the module descriptor.

craigmcnally:
Also:
* This is a breaking change - the version should be bumped in module descriptor interface, raml.  
* Don't forget to cleanup permissions.
* Adding a note in the Q2 release notes would be a good idea (if you haven't already) indicating that there are permission changes - Since FOLIO doesn't currently handle permission migration the only thing we can do is make sure folks are aware things changed.  If they want to clean up or make adjustments they will have to do so manually.

PR #71, opened 2020-05-28
Modoaipmh-108 Make deleted instance records support configurable

KITSViktorSkibnev:
## Purpose

Filter records from source-record-storage by the fields deleted = true and MARC Leader 05 is "d", "s" or "x" in java classes.

## Approach
To do this, in the code, I check all the records according to the above criteria (fields "deleted = true" and MARC Leader 05 is "d", "s" or "x" or one is completed) and if the record is marked as deleted and repository.deletedRecords to be configured as "persistent" or "transient ", then I add status="deleted" in header and do not add metadata. In this case i make request to "/source-storage/records" path for get records with "deleted = true" because when i make request to "/source-storage/sourceRecords" (old behavior) i get only records with "deleted = false" and I can’t filter records by this field. If repository.deletedRecords to be configured as "no" old behavior is works. I also check the suppressDiscovery config setting and if suppressDiscovery=true and record marked as deleted, record present in oai-pmh response. In another case (suppressDiscovery=true and record not marked as deleted) record is absent in oai-pmh response.

### LINKS
[https://issues.folio.org/browse/MODOAIPMH-108](url)


src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -315,6 +354,25 @@ private String getUntilDate(Request request, String from) {
     }
   }
 
+  protected boolean filterInstance(Request request, JsonObject instance){
+    if (!isDeletedRecordsEnabled(request, REPOSITORY_DELETED_RECORDS)){
+      return !storageHelper.isRecordMarkAsDeleted(instance);
+    } else {
+      Map<String, String> okapiHeaders = request.getOkapiHeaders();
+      return getBooleanProperty(okapiHeaders, REPOSITORY_SUPPRESSED_RECORDS_PROCESSING)
```
IlliaDaliek, line 369: Could you create please the boolean variable here "shouldProcessSuppressedRecords" and assign to it value of the setting and use such variable instead of "getBooleanProperty (okapiHeaders, REPOSITORY_SUPPRESSED_RECORDS_PROCESSING)" which takes a lot of space and make code less readable.

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -66,6 +75,18 @@
    */
   protected StorageHelper storageHelper = StorageHelper.getInstance();
 
+  /**
+   * Creates basic {@link OAIPMH} with ResponseDate and Request details
+   * @param request {@link Request}
+   * @return basic {@link OAIPMH}
+   */
+  protected OAIPMH buildBaseResponse(Request request) {
```
IlliaDaliek, line 84: Please remove this. This logic was extracted to ResponseHelper::buildBaseOaipmhResponse. Any response building logic should bases on ResponseHelper class methods.

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -257,6 +278,19 @@ protected String getIdentifier(String identifierPrefix, String storageId) {
     return identifierPrefix + storageId;
   }
 
+  /**
+   * The error codes required to define the http code to be returned in the http response
+   * @param oai OAIPMH response with errors
+   * @return set of error codes
+   */
+  protected Set<OAIPMHerrorcodeType> getErrorCodes(OAIPMH oai) {
```
IlliaDaliek, line 293: The same is here. This was extracted to ResponseHelper as well.

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -276,7 +310,12 @@ protected ResumptionTokenType buildResumptionToken(Request request, JsonArray in
       Map<String, String> extraParams = new HashMap<>();
       extraParams.put(TOTAL_RECORDS_PARAM, String.valueOf(totalRecords));
       extraParams.put(OFFSET_PARAM, String.valueOf(newOffset));
-      String nextRecordId = storageHelper.getRecordId((JsonObject) instances.remove(instances.size() - 1));
+      String nextRecordId;
+      if (isDeletedRecordsEnabled(request, REPOSITORY_DELETED_RECORDS)){
+        nextRecordId = storageHelper.getId((JsonObject) instances.remove(instances.size() - 1));
```
IlliaDaliek, line 322: It is a bit confusing what happens here. May be would be better to move this to a simple short private method with clear name which will describe what it actually does and in such way we would keep the same level of abstraction within current method.

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -315,6 +354,25 @@ private String getUntilDate(Request request, String from) {
     }
   }
 
+  protected boolean filterInstance(Request request, JsonObject instance){
```
IlliaDaliek: Could you please add javadoc and describe what happens here. It would be useful.

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -332,9 +390,11 @@ private String getUntilDate(Request request, String from) {
    */
   protected boolean canResumeRequestSequence(Request request, Integer totalRecords, JsonArray instances) {
     Integer prevTotalRecords = request.getTotalRecords();
+    String recordId = isDeletedRecordsEnabled(request, REPOSITORY_DELETED_RECORDS)
+      ? storageHelper.getId(instances.getJsonObject(0)) : storageHelper.getRecordId(instances.getJsonObject(0));
```
IlliaDaliek: Move '0' to constants please with clear naming what it means.

src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
```diff
@@ -57,9 +59,10 @@
       }
 
       HttpClientInterface httpClient = getOkapiClient(request.getOkapiHeaders());
+      final String instanceEndpoint = storageHelper.buildRecordsEndpoint(request,isDeletedRecordsEnabled(request, REPOSITORY_DELETED_RECORDS));
```
IlliaDaliek: It seems to me that you haven't formatted the code properly. There is no whitespace after the comma. Such lines make it a bit hard to read the code.

src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
```diff
@@ -118,10 +126,19 @@ private OAIPMH buildListIdentifiers(Request request, Response instancesResponse)
       instances.stream()
         .map(object -> (JsonObject)object)
         .filter(instance -> StringUtils.isNotEmpty(storageHelper.getIdentifierId(instance)))
-        .map(instance -> populateHeader(identifierPrefix, instance))
+        .filter(instance -> filterInstance(request, instance))
+        .map(instance -> addHeader(identifierPrefix, request, instance))
         .forEach(identifiers::withHeaders);
+
```
IlliaDaliek, line 133: Remove one empty line please.

src/main/java/org/folio/oaipmh/helpers/storage/InventoryStorageHelper.java
```diff
@@ -42,7 +42,7 @@ public String getIdentifierId(final JsonObject entry) {
   }
 
   @Override
-  public String buildRecordsEndpoint(Request request) throws UnsupportedEncodingException {
+  public String buildRecordsEndpoint(Request request, boolean recordsPath) throws UnsupportedEncodingException {
```
IlliaDaliek, line 45: For boolean variable the recordsPath is not the appropriate name. So, for me it's confusing to get what this variable means here from logic side. Rename it like isRecordPath.

src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationUtil.java
```diff
@@ -112,4 +114,11 @@ public static boolean getBooleanProperty(Map<String, String> okapiHeaders, Strin
     return parseBoolean(defaultValue);
   }
 
+  public static boolean isDeletedRecordsEnabled(Request request, String name){
+    String tenant = TenantTool.tenantId(request.getOkapiHeaders());
+    String propertyName = getProperty(tenant, name);
+
+    return (DeletedRecordType.fromValue(propertyName)).equals(DeletedRecordType.PERSISTENT)
```
IlliaDaliek, line 121: So, if there is no enum instance for 'propertyName' then exception will be throwed. Please wrap it with try catch and wihin the catch block  set up the default value for setting. You can watch the default value within config file into resource dirrectory

src/main/java/org/folio/oaipmh/helpers/storage/SourceRecordStorageHelper.java
```diff
@@ -91,4 +102,36 @@ public String getRecordByIdEndpoint(String id) {
   public boolean getSuppressedFromDiscovery(final JsonObject entry) {
     return entry.getJsonObject(ADDITIONAL_INFO).getBoolean(SUPPRESS_DISCOVERY);
   }
+
+  private String getLeaderValue(JsonObject entry){
+    return Optional.ofNullable(entry.getJsonObject(PARSED_RECORD))
+      .map(record -> record.getJsonObject(CONTENT))
+      .map(content -> content.getString("leader"))
+      .orElse(null);
```
IlliaDaliek, line 112: It is a bad practice of returning the null. Empty string would be much better.

craigmcnally:
> In this case i make request to "/source-storage/records" path for get records with "deleted = true" because when i make request to "/source-storage/sourceRecords" (old behavior) i get only records with "deleted = false"

Is this a bug, or expected behavior?

KITSViktorSkibnev:
Hi @craigmcnally ! It's expected behavior because endpoint "/source-storage/sourceRecords" return only records with "deleted = false" but for filtering records by  fields "deleted = true" and MARC Leader 05 is "d", "s" or "x" we need records with "deleted = true" and i use new endpoind.

craigmcnally:
An out-of-band conversation revealed that this is a temporary workaround and that another solution would be adopted "later on".

I'm abstaining from approval of this as I don't have all the details.  Unanswered questions I have are:
* Is the described behavior of /sourceRecords **desired** behavior or a bug?
* What is the differences (functional, performance, otherwise) between the /records and /sourceRecords endpoints.

mrdpopov:
@craigmcnally I believe a misunderstanding happened here. 
The endpoint we’re using (/records instead of /sourceRecords) isn’t a workaround, but the only solution we could go with. The main difference of these two - /sourceRecords filters out records marked as “deleted”, but /records doesn’t do it. 

As far as I know, described behavior of /sourceRecords is a desired behavior as of now, since there is additional request parameter “deleted”, which allows us to search only for deleted records. However it isn’t convenient for us, as we need to get both – deleted and not deleted records in one response – and only /records endpoint allows us to do that.

One more thing, that might seem confusing, SRS is in process of refactoring and extending functionality now, so we asked its dev team to look into additional filtering cases to be performed on the SRS level and be available via its API (corresponding ticket is https://issues.folio.org/browse/MODSOURCE-133). With these new abilities we’ll move some logic (e.g. checking leader 05 field) from oai-pmh java code to SRS querying, which should improve performance (see https://issues.folio.org/browse/MODOAIPMH-136 for details).

PR #88, opened 2020-06-24
Modoaipmh 176

IlliaDaliek:
### PURPOSE
Fix issue with posting configurations to mod-ocnfiguration while running TenantAPI & use ConfigurationCLient instead of HttpClientInterface

### DIFF
--- a/src/main/java/org/folio/rest/impl/ModTenantAPI.java
+++ b/src/main/java/org/folio/rest/impl/ModTenantAPI.java
@@ -41,7 +41,7 @@ public class ModTenantAPI extends TenantAPI {
   private static final String X_OKAPI_TENANT = "x-okapi-tenant";
   private static final String X_OKAPI_TOKEN = "x-okapi-token";
   private static final String CONFIG_DIR_PATH = "config";
-  private static final String QUERY = "module==OAIPMH+and+configName==%s";
+  private static final String QUERY = "module==OAIPMH and configName==%s";
   private static final String MODULE_NAME = "OAIPMH";
   private static final int CONFIG_JSON_BODY = 0;
   private ConfigurationHelper configurationHelper = ConfigurationHelper.getInstance();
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -1,142 +1,157 @@
...
 
+  private Future processConfigurationByConfigName(String configName, Map<String, String> headers) {
+    Promise promise = Promise.promise();
     String okapiUrl = headers.get(X_OKAPI_URL);
     String tenant = headers.get(X_OKAPI_TENANT);
+    String token = headers.get(X_OKAPI_TOKEN);
 
-    HttpClientInterface httpClient = HttpClientFactory.getHttpClient(okapiUrl, tenant, true);
-    List<CompletableFuture<Void>> completableFutures = new ArrayList<>();
+    ConfigurationsClient client = new ConfigurationsClient(okapiUrl, tenant, token);
```
mrdpopov, line 78: The client is created 3 times. Can you create it once and pass it to this method?

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -1,142 +1,157 @@
...
+  private static final String MODULE_NAME = "OAIPMH";
   private static final int CONFIG_JSON_BODY = 0;
   private ConfigurationHelper configurationHelper = ConfigurationHelper.getInstance();
 
   @Override
   public void postTenant(final TenantAttributes entity, final Map<String, String> headers,
       final Handler<AsyncResult<Response>> handlers, final Context context) {
-    loadConfigData(headers, context).thenAccept(response -> handlers.handle(Future.succeededFuture(response)))
-      .exceptionally(handleError(handlers));
+    loadConfigData(headers, handlers);
```
mrdpopov, line 51: Please inline this method.

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -1,142 +1,157 @@
...
+    List<Future> futures = new ArrayList<>();
+
+    configsSet.forEach(configName -> futures.add(processConfigurationByConfigName(configName, headers)));
+    CompositeFuture.all(futures)
+      .onComplete(future -> {
+        String message;
+        if (future.succeeded()) {
+          message = "Configurations has been set up successfully";
+        } else {
+          message = future.cause().getMessage();
```
mrdpopov, line 66: Possible nullpointer

mrdpopov:
LFTM

PR #89, opened 2020-06-24
Update Requests to SRS for v4

mrdpopov:
# Purpose

Refactoring to support SRS for v4


mrdpopov:
Tests to be fixed in the following commits

PR #92, opened 2020-06-25
Use log4j2.properties

mrdpopov:
## Purpose

Use log4j2.properties instead of log4j.properties


craigmcnally:
> Use log4j.properties instead of log4j2.properties
You got it backwards here :)

mrdpopov:
Done

PR #94, opened 2020-06-26
Update to SRS for v4

mrdpopov:
## Purpose

Update oai-pmh to SRS for v4


src/main/java/org/folio/oaipmh/helpers/storage/StorageHelper.java
```diff
@@ -21,8 +17,6 @@ static StorageHelper getInstance() {
     String repositoryType = System.getProperty(REPOSITORY_STORAGE, SOURCE_RECORD_STORAGE);
     if (SOURCE_RECORD_STORAGE.equals(repositoryType)) {
       return new SourceRecordStorageHelper();
-    } else if (INVENTORY_STORAGE.equals(repositoryType)) {
```
craigmcnally, line 24: I'm thinking that maybe it would be better to default to SRS and log a warning instead of removing this completely.  Doing it this way means this is a breaking change and anyone using this option would need to explicitly change it to get the module running again.

Either way, we should probably note this in the goldenrod release notes.  I don't *think* anyone is using this option, but who knows.

craigmcnally: And really, if we're only going to support this one storage, we can probably clean a lot of this up - Maybe track this in a story tagged as tech-debt?

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -118,8 +119,8 @@ private void postConfig(ConfigurationsClient client, String configName, Promise<
     try {
       client.postConfigurationsEntries(null, config, resp -> {
         if (resp.statusCode() != 201) {
-          logger.error("Invalid response from mod-configuration. Cannot post config '{}': {} {}", configName, resp.statusCode(),
-              resp.statusMessage());
+          logger.error(String.format("Invalid response from mod-configuration. Cannot post config '%s'." +
```
craigmcnally, line 122: Probably better to stick with slf4j concatenation... see https://stackoverflow.com/questions/10555409/logger-slf4j-advantages-of-formatting-with-instead-of-string-concatenation

mrdpopov: It was not working, that's why I changed it. I'll double check.

src/test/java/org/folio/oaipmh/helpers/StorageHelperTest.java
```diff
@@ -1,250 +1,166 @@
...
+    System.setProperty(REPOSITORY_STORAGE, storageType);
+    return StorageHelper.getInstance();
+  }
+
+  private String getDirPath(String storageType) {
+    return SOURCE_RECORD_STORAGE.equals(storageType) ? SOURCE_STORAGE_RECORD_PATH : "TODO CHANGE";
+  }
+
+  private String getDirPathWithPathRecordsInSRS(String storageType) {
+    return SOURCE_RECORD_STORAGE.equals(storageType) ? STORAGE_RECORD_PATH : "TODO CHANGE";
```
craigmcnally, line 140: I **_think_** TODO CHANGE here suggests that you plan on cleaning up the ability to specify the storage type (see earlier comment)... +1

src/main/java/org/folio/oaipmh/helpers/storage/StorageHelper.java
mrdpopov: This inventory-storage has been removed because it is a) not used from the business requirements perspective b) we have implemented another means of communicating with inventory c) this class was hard to maintain, was obsolete and entailed writing/supporting tests  for it.
So for metadata prefix marc21_withholdings we use another method of getting instances from inventory, inventory-view. And for all other metadata prefixes we do not use inventory at all.

craigmcnally: oh, interesting - so the ability to do this is temporarily removed, but slated to be re-added in Q3?

mrdpopov: It will be a completely different mechanism that talks to inventory in Q3, see https://github.com/folio-org/mod-oai-pmh/blob/master/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java#L292

src/test/java/org/folio/oaipmh/helpers/StorageHelperTest.java
mrdpopov: Well, not it is not possible for end-user to specify the storage type, we will be adding this feature in Q3. Sorry for not cleaning this up.

PR #102, opened 2020-07-08
Suppress holdings and items records from discovery

mrdpopov:
## Purpose

- Send 0 or 1 flag for suppressed records in 856(electronic access) and 952(location) fields
- Fill 856 $t field (which comes from Inventory) with suppress from discovery info if this field exists in marc record
- Fix 952$d(location name) and 952$n (copy number)

Jira: https://issues.folio.org/browse/MODOAIPMH-107

## Notes
According to to the latest requirements, 999 field should be taken from SRS record without changes, it shouldn't be recreated in mod-oai-pmh. The value to be transferred depending on the Suppressed records processing setting.


src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
```diff
@@ -160,11 +163,11 @@ private Response buildRecordsResponse(Context ctx, Request request,
           }
           // Some repositories like SRS can return record source data along with other info
           String source = storageHelper.getInstanceRecordSource(instance);
-          if (source != null) {
-            source = metadataManager.updateMetadataSourceWithDiscoverySuppressedDataIfNecessary(source, instance, request);
```
oleksandr-yatsenko1, line 164: In old implentation "discovery suppressed data" was processed even without record header's status checking. Now we process "suppredd discovery data" in couble wth rest of operation related to record filling if there is no header status to. Does it intended by business case?

mrdpopov: It is just reordering and optimizing ifs

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -209,35 +241,56 @@ private void addSubFieldGroup(Map<String, Object> effectiveLocationSubFields, Js
...
+
+
   private Optional<JsonObject> getGeneralInfoDataField(JsonArray fields) {
     return fields.stream()
       .map(obj -> (JsonObject) obj)
       .filter(generalInfoFieldPredicate)
       .findFirst();
   }
 
+  private Optional<JsonObject> getElectronicAccessField(JsonArray fields) {
```
oleksandr-yatsenko1, line 281: Taking into account that JsaonArray is raw typed, is it fully safe to use implicit type casting here? Possibly it would be better to do slemting like iteration over stream range and getting JsonObject by index: 

IntStream.range(0, fields.size()).mapToObj(idx -> fields.getJsonObject(idx)).filter(electronicAccessPredicate).findFirst();

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -116,9 +128,17 @@ public JsonObject populateMetadataWithItemsData(JsonObject srsInstance, JsonObje
...
+   * @param suppressedRecordsProcessing - include suppressed flag in 952 field?
    */
-  private void updateFieldsWithItemEffectiveLocationField(JsonObject itemData, List<Object> marcRecordFields) {
+  private void updateFieldsWithItemEffectiveLocationField(JsonObject itemData,
+                                                          List<Object> marcRecordFields,
+                                                          boolean suppressedRecordsProcessing) {
     Map<String, Object> effectiveLocationSubFields = constructEffectiveLocationSubFieldsMap(itemData);
+    if (suppressedRecordsProcessing) {
+      final Boolean itemSuppressedFromDiscovery = itemData.getBoolean(INVENTORY_SUPPRESS_DISCOVERY_FIELD);
+      int subFieldValue = itemSuppressedFromDiscovery == null || !itemSuppressedFromDiscovery ? 0 : 1;
```
oleksandr-yatsenko1, line 139: Taking into account that we are already used "apache common lang3" this confditions can be replaced by BooleanUtils.isFalse(). How fo you think, does it make sence?

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -134,15 +154,22 @@ private void updateFieldsWithItemEffectiveLocationField(JsonObject itemData, Lis
...
     JsonArray electronicAccessArray = itemData.getJsonArray(ELECTRONIC_ACCESS);
     if (Objects.nonNull(electronicAccessArray)) {
       electronicAccessArray.forEach(electronicAccess -> {
         Map<String, Object> electronicAccessSubFields = constructElectronicAccessSubFieldsMap((JsonObject) electronicAccess);
         FieldBuilder fieldBuilder = new FieldBuilder();
-
         List<String> indicators = resolveIndicatorsValue((JsonObject) electronicAccess);
+        if (suppressedRecordsProcessing) {
+          final Boolean itemSuppressedFromDiscovery = itemData.getBoolean(INVENTORY_SUPPRESS_DISCOVERY_FIELD);
+          int subFieldValue = itemSuppressedFromDiscovery == null || !itemSuppressedFromDiscovery ? 0 : 1;
```
oleksandr-yatsenko1, line 170: The same here: taking into account that we are already used "apache common lang3" this confditions can be replaced by BooleanUtils.isFalse(). How fo you think, does it make sence?

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -160,7 +187,7 @@ private void updateFieldsWithItemElectronicAccessField(JsonObject itemData, List
     String name = electronicAccess.getString(NAME);
     String key = StringUtils.isNotEmpty(name) ? name : EMPTY;
     String indicatorsInString = indicatorsMap.get(key);
-    if(indicatorsInString!=null) {
+    if (indicatorsInString != null) {
```
oleksandr-yatsenko1, line 190: Possibly this can be relpaces by simpler case to avoid conditional operator usage, I man something like this:

return Arrays.asList(Optional.ofNullable(indicatorsInString ).orElse(StringUtils.EMPTY).split(","));

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -175,11 +202,16 @@ private void updateFieldsWithItemElectronicAccessField(JsonObject itemData, List
     addSubFieldGroup(effectiveLocationSubFields, locationGroup, EffectiveLocationSubFields.getLocationValues());
     addSubFieldGroup(effectiveLocationSubFields, callNumberGroup, EffectiveLocationSubFields.getCallNumberValues());
     addSubFieldGroup(effectiveLocationSubFields, itemData, EffectiveLocationSubFields.getSimpleValues());
+    //Map location name, which changed paths in json, to 952$d
+    String subFieldValue = itemData.getJsonObject(LOCATION).getString(NAME);
+    if (isNotEmpty(subFieldValue)) {
```
oleksandr-yatsenko1, line 207: Possibly "isBlank" would cover wider data checking, does it make sence to use it instead ?

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -175,11 +202,16 @@ private void updateFieldsWithItemElectronicAccessField(JsonObject itemData, List
     addSubFieldGroup(effectiveLocationSubFields, locationGroup, EffectiveLocationSubFields.getLocationValues());
     addSubFieldGroup(effectiveLocationSubFields, callNumberGroup, EffectiveLocationSubFields.getCallNumberValues());
     addSubFieldGroup(effectiveLocationSubFields, itemData, EffectiveLocationSubFields.getSimpleValues());
+    //Map location name, which changed paths in json, to 952$d
+    String subFieldValue = itemData.getJsonObject(LOCATION).getString(NAME);
```
oleksandr-yatsenko1, line 206: It looks not safe, NPE can happened. Possibly this can be wrapped with something like this:
Optional.ofNullable(itemData.getJsonObject(LOCATION)).orElse(new JsonObject()).getString(NAME) ?

craigmcnally:
I find it very helpful to have a link to the JIRA in the PR description.  I'm not saying you have to, but please consider doing this in the future

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: I think it is overcomplication, we'll have to change everywhere else.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: Done.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -71,6 +71,16 @@ private RecordMetadataManager() {
       }
       return false;
     };
+
+    electronicAccessPredicate = jsonObject -> {
+      if (jsonObject.containsKey(ELECTRONIC_ACCESS_FILED_TAG_NUMBER)) {
+        JsonObject dataFieldContent = jsonObject.getJsonObject(ELECTRONIC_ACCESS_FILED_TAG_NUMBER);
+        String firstIndicator = dataFieldContent.getString(FIRST_INDICATOR);
+        String secondIndicator = dataFieldContent.getString(SECOND_INDICATOR);
+        return StringUtils.isNotEmpty(firstIndicator) && StringUtils.isNotEmpty(secondIndicator);
```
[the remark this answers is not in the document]
mrdpopov: done.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: Done.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: done.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: done.

src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
mrdpopov: All existing code can't be fixed in one pull request, can it?

src/test/resources/metadata-manager/inventory_instance_with_2_items.json
```diff
@@ -61,6 +62,7 @@
         },
         "enumeration": "Enumeration 1_2_1",
         "materialType": "book",
+        "copyNumber": "copy number",
```
craigmcnally, line 65: This is on the inventory guys not you... seems silly to all it copyNumber when it's not a number

src/test/resources/metadata-manager/srs_instance_with_electronic_access.json
```diff
@@ -432,6 +452,20 @@
...
+        {
+          "999": {
+            "ind1": "f",
+            "ind2": "f",
+            "subfields": [
+              {
+                "s": "3c4ae3f3-b460-4a89-a2f9-78ce3145e4fc"
+              },
+              {
+                "i": "00000000-0000-4000-a000-000000000000"
```
craigmcnally, line 465: What's the significance of this?  Is there some hidden/inferred meaning here, or could this just as easily be any old UUID?

src/test/resources/metadata-manager/inventory_instance_with_2_items.json
mrdpopov: It was just a string for testing purposes

src/test/resources/metadata-manager/srs_instance_with_electronic_access.json
mrdpopov: No extra significance, this was supposed to be an instance with electronic access, all other fields have been copied from the existing file

PR #105, opened 2020-07-10
MODOAIPMH-191 Connect the database to mod-oai-pmh for sets functionality

IlliaDaliek:
### PURPOSE
Connect the databse to the module and create tables for sets.
https://issues.folio.org/browse/MODOAIPMH-191


src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -99,8 +125,9 @@ private void handleModConfigurationGetResponse(HttpClientResponse response, Conf
       JsonObject jsonConfig = body.toJsonObject();
       JsonArray configs = jsonConfig.getJsonArray(CONFIGS);
       if (configs.isEmpty()) {
-        logger.info(String.format("Configuration group with configName %s isn't exist. " +
-          "Posting default configs for '%s' configuration group", MODULE_NAME, configName));
+        logger.info(String.format(
```
mrdpopov, line 128: Please use {} from log4j instead of String format

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -165,6 +193,45 @@ private String getConfigValue(String configName) {
     return configEntryValueField.encode();
   }
 
+  private Future<String> initDatabase(Map<String, String> okapiHeaders, Vertx vertx) {
+    Promise<String> promise = Promise.promise();
+    String tenantId = okapiHeaders.get(OKAPI_TENANT);
+    vertx.executeBlocking(blockingFeature -> {
+      logger.info("KEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEK");
```
MikhailFokanov, line 200: What is it?

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -165,6 +193,45 @@ private String getConfigValue(String configName) {
     return configEntryValueField.encode();
   }
 
+  private Future<String> initDatabase(Map<String, String> okapiHeaders, Vertx vertx) {
+    Promise<String> promise = Promise.promise();
+    String tenantId = okapiHeaders.get(OKAPI_TENANT);
+    vertx.executeBlocking(blockingFeature -> {
```
MikhailFokanov, line 199: Why it should be "executeBlocking"?

IlliaDaliek: > Why it should be "executeBlocking"?

Writing this code i've relied on code into SRS and mod-pub-sub and they use executeBLocking, so i've used the same approach.

src/main/java/org/folio/rest/impl/ModTenantAPI.java
IlliaDaliek: > What is it?
My custom mark for logging the especially interested code part for me. Just forgot to remove. Already removed.

src/main/java/org/folio/rest/impl/ModTenantAPI.java
IlliaDaliek: > Please use {} from log4j instead of String format

Good note, thank you

PR #106, opened 2020-07-15
MODOAIPMH-203: Enable Spring DI for mod-oai-pmh

IlliaDaliek:
### PURPOSE
Enable Spring DI for mod-oai-pmh.
https://issues.folio.org/browse/MODOAIPMH-203


src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -46,7 +48,12 @@
   private static final String QUERY = "module==OAIPMH and configName==%s";
   private static final String MODULE_NAME = "OAIPMH";
   private static final int CONFIG_JSON_BODY = 0;
-  private ConfigurationHelper configurationHelper = ConfigurationHelper.getInstance();
+  @Autowired
```
mrdpopov, line 51: Please do not use field injection

mrdpopov:
LGTM

PR #117, opened 2020-08-06
MODOAIPMH-204: (mod-oai-pmh) release 3.1.0

IlliaDaliek:
### PURPOSE
Release oia-pmh 3.1.0 module

The main functionality of this release is MODOAIPMH-202. Additionally, release build failure when building locally has been 
fixed: Jira: MODOAIPMH-213
As well the databse has been introduced (MODOAIPMH-191) and the spring DI support has been enabled too (MODOAIPMH-203).


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-oai-pmh</artifactId>
-  <version>3.0.4</version>
+  <version>3.2.0-SNAPSHOT</version>
```
craigmcnally, line 5: If you're releasing off this branch it should be 3.1.0

pom.xml
```diff
@@ -18,7 +18,7 @@
     <url>https://github.com/folio-org/mod-oai-pmh.git</url>
     <connection>scm:git:git://github.com/folio-org/mod-oai-pmh.git</connection>
     <developerConnection>scm:git:git@github.com:folio-org/mod-oai-pmh.git</developerConnection>
-    <tag>v3.0.4</tag>
+    <tag>HEAD</tag>
```
craigmcnally, line 21: see above

pom.xml
craigmcnally: My guess is that you already created the tag - unfortunately, that means if you need to make any adjustments - in reaction to PR comments for instance, you'll need to create a new tag and bump the version again.  The CI system doesn't handle moving the tag to a new commit very well.

pom.xml
```diff
@@ -131,6 +131,12 @@
       <artifactId>domain-models-runtime</artifactId>
       <version>${raml-module-builder.version}</version>
     </dependency>
+    <dependency>
```
craigmcnally, line 134: I know there was a separate PR, but was there a separate JIRA for the introduction of DI support?  Either way, it might be worth mentioning this in the release notes

pom.xml
```diff
@@ -573,7 +591,6 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.22.2</version>
         <configuration>
-          <argLine>-Xmx1024m -XX:MaxPermSize=256m</argLine>
```
craigmcnally, line 576: was there a reason to remove this in a release PR?

pom.xml
craigmcnally: wait - is this just now being added as part of the release PR?  I thought it was added earlier in another PR?

pom.xml
IlliaDaliek: You are right, yes, it was by mistake that we are not goint to merge it to mainline but to long-live branch b3.0. I have fixed it. Thank you.

pom.xml
IlliaDaliek: See my comment above, i've fixed it.

src/test/resources/metadata-manager/inventory_instance_with_2_items.json
```diff
@@ -1,6 +1,6 @@
 {
   "instanceid": "2f7f23be-81a6-4227-85ba-7c1c81272ee1",
-  "instanceupdateddate": "2020-06-01T10:56:10.118Z",
+  "updateddate": "2020-06-01T10:56:10.118Z",
```
craigmcnally, line 3: I must be missing some detail here... why aren't these camelCase?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -0,0 +1 @@
+{}
```
craigmcnally, line 1: I'm guessing this is needed and not superfluous.  Is it part of the liquidbase work?

mrdpopov: yes

src/test/resources/metadata-manager/inventory_instance_with_2_items.json
IlliaDaliek: We have a database view which returns data in such manner, that's why these fields are named in such way.

craigmcnally: > We have a database view which returns data in such manner
why?

mrdpopov: Well, it's in another module(which is already released), we have to integrate with it. If you think this should be changed, I suggest we do it separately.

craigmcnally: What I'm seeing in the schemas is that these fields are camelCase... Am I looking at the wrong APIs?

https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/oai-pmh-view.html#

craigmcnally: correction - I see that there are a few fields which aren't camelCase... I'm really curious why this was done as it's inconsistent, not only within the schema but across all of FOLIO

craigmcnally: I'm looking at this PR:  https://github.com/folio-org/mod-inventory-storage/pull/473 and the schemas use camelCase, but when I actually make calls to the API (folio-testing) they're all lowercase.

mrdpopov: We're using
https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/oai-pmh-view.html#oai_pmh_view_updatedinstanceids_get
and
https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/oai-pmh-view.html#oai_pmh_view_enrichedinstances_post

mrdpopov: Yes, when a call is made to API, the values are in lower case

craigmcnally: right... and look at the response schema for those APIs... they have "instanceId" (camelCase).  but when you call the actual API you mysteriously get "instanceid" (all lowercase).  I'm confused what's going on here

mrdpopov:
> 
> 
> Approving with the stipulation that the naming convention/case oddities are investigated and fixed (see comments above).
> 
> @IlliaDaliek / @mrdpopov please create a JIRA to track this and add a link here

https://issues.folio.org/browse/MODINVSTOR-553

PR #118, opened 2020-08-06
MODOAIPMH-213: Maven build test phase fails 

IlliaDaliek:
### PURPOSE
Fix maven build and resolve issue with unnecessary logging.
https://issues.folio.org/browse/MODOAIPMH-213

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,12 @@
       <artifactId>folio-liquibase-util</artifactId>
       <version>1.0.0</version>
       <type>jar</type>
+      <exclusions>
+        <exclusion>
+          <groupId>ch.qos.logback</groupId>
+          <artifactId>logback-classic</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <!-- Jaxb dependencies -->
     <dependency>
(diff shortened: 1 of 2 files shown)


craigmcnally:
I see the error in the JIRA, but what was the actual issue here?  Was there a conflict w/ the logback dependency?  Or was it the JVM args that were causing problems?  Or was the schema creation part that needed to be adjusted (if not exists)?

IlliaDaliek:
> I see the error in the JIRA, but what was the actual issue here? Was there a conflict w/ the logback dependency? Or was it the JVM args that were causing problems? Or was the schema creation part that needed to be adjusted (if not exists)?

The main problem that prevents normal run maven release commands was an endless logging to console (see image has been attached). Such log output was caused by logback library that loaded by org-folio-liquibase dependency. So removing this dependency we've solved the problem and released the module. 
As well regarding your question due to json lower case fields in https://github.com/folio-org/mod-oai-pmh/pull/117, we've created a JIRA ticket (https://issues.folio.org/browse/MODINVSTOR-553) to mod-invenotry and started investigation of the problem.
logging

PR #119, opened 2020-08-10
MODOAIPMH-218: Rename sets endpoints 

IlliaDaliek:
### PURPOSE
Renaming of set endpoints due to current inconsistencies with common approach for writing the raml files structure on project which is going to bring a problems in furthet CRUD API  implementation for "sets" and "filtering conditions".
https://issues.folio.org/browse/MODOAIPMH-218

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -24,23 +24,23 @@
         },
         {
           "methods": ["GET"],
-          "pathPattern": "/oai-pmh/set/{id}",
-          "permissionsRequired": ["oai-pmh.set.item.get"]
+          "pathPattern": "/oai-pmh/sets/{id}",
+          "permissionsRequired": ["oai-pmh.sets.item.get"]
         },
         {
           "methods": ["PUT"],
-          "pathPattern": "/oai-pmh/set/{id}",
-          "permissionsRequired": ["oai-pmh.set.item.put"]
+          "pathPattern": "/oai-pmh/sets/{id}",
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -633,7 +633,6 @@
       <plugin>
         <groupId>com.gitlab.janecekpetr</groupId>
         <artifactId>embedded-postgresql-maven-plugin</artifactId>
-        <version>0.1.0</version>
```
mreno-EBSCO, line 636: I am curious why the version was removed.

IlliaDaliek: > I am curious why the version was removed.

Cause we have had a problems with building branch on rancher and by removing the version we've solved the problem and this change doesn't affect anything, so it's okay here. And removing the version the maven automatically uses the last one.

mreno-EBSCO: I disagree that this is OK. This breaks build reproducibility. I suggest we discover why building branches on rancher is broken rather than making our POM files ambiguous.

mrdpopov: Ok let's put it back.

IlliaDaliek: > I disagree that this is OK. This breaks build reproducibility. I suggest we discover why building branches on rancher is broken rather than making our POM files ambiguous.

I've returned the version.

IlliaDaliek: > I disagree that this is OK. This breaks build reproducibility. I suggest we discover why building branches on rancher is broken rather than making our POM files ambiguous.

@mreno-EBSCO could you approve the code review please if it's okay now on your opinion.

PR #121, opened 2020-08-12
MODOAIPMH-200: Implement the endpoint for getting list of sets

IlliaDaliek:
### PURPOSE
Implement new endpoint /oai-pmh/sets which is responsible for getting a list of sets and make it pageable (supports offset and limit query parameters).
As well increase postgres connection pool size.
https://issues.folio.org/browse/MODOAIPMH-200


src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
```diff
@@ -138,7 +150,7 @@ private SetRecord toDatabaseSetRecord(Set set) {
     if (isNotEmpty(set.getName())) {
       dbRecord.setName(set.getName());
     }
-    if (isNotEmpty(set.getDescription())) {
+    if (Objects.nonNull(set.getDescription())) {
```
craigmcnally, line 153: Does this mean that there's no way to remove a description?

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
```diff
@@ -120,14 +118,28 @@ public SetDaoImpl(final PostgresClientFactory postgresClientFactory) {
...
+      .collect(Collectors.toList());
+    return new SetCollection().withSets(list);
+  }
+
   private void prepareSetMetadata(Set entry, String userId, InsertType insertType) {
-    if (insertType.equals(InsertType.UPDATE)) {
-      entry.setUpdatedDate(from(Instant.now()));
-      entry.setUpdatedByUserId(userId);
-    } else {
+    if (insertType.equals(InsertType.INSERT)) {
```
craigmcnally, line 137: Why do we need to manually manage metadata?  Can't the database/RMB do that for us?

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
```diff
@@ -91,12 +95,32 @@ public void deleteOaiPmhSetsById(String id, String lang, Map<String, String> oka
...
     vertxContext.runOnContext(v -> {
       try {
+        logger.info("Delete set by id with id: '{}'", id);
         setService.deleteSetById(id, getTenantId(okapiHeaders))
           .map(DeleteOaiPmhSetsByIdResponse.respond204())
           .map(Response.class::cast)
           .otherwise(ExceptionHelper::mapExceptionToResponse)
           .onComplete(asyncResultHandler);
       } catch (Exception e) {
+        logger.error("Error occurred while posting set with id: '{}'. Message: {}. Exception: {}", id, e.getMessage(), e);
```
craigmcnally, line 105: not posting - deleting

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: > Does this mean that there's no way to remove a description?

No, at opposite its a fix for what you've said) before we hadn't the possibility to remove description.

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
IlliaDaliek: > not posting - deleting

I don't think so, there are will be changes relatively to metadata soon, it will be leaded to the common metadata strucutre as it used on the project. We don't want to save it as nested json object to database (I mean the json field type) cause we have to implement seraching mechanism which uses metadata fields and this won't be optimize to parse this json within searching logic.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: > Why do we need to manually manage metadata? Can't the database/RMB do that for us?

I don't think so, there are will be changes relatively to metadata soon, it will be leaded to the common metadata strucutre as it used on the project. We don't want to save it as nested json object to database (I mean the json field type) cause we have to implement seraching mechanism which uses metadata fields and this won't be optimize to parse this json within searching logic.

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
IlliaDaliek: Thank you for note. Corrected

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: > Why do we need to manually manage metadata? Can't the database/RMB do that for us?

As well we don't use rmb to manage our database records, we use another approach (liquibase + jooq) in compare with other folio modules.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
craigmcnally: if I start with:

```
{
  "id": "12345",
  "name": "foobar",
  "description": "hello"
}
```

and put:
```
{
  "id": "12345",
  "name": "foobar"
}
```

Then description will be null, the condition will evaluate to false and the description will remain as `"hello"`, right?  I guess I'm not sure why you need these conditionals at all...

IlliaDaliek: > if I start with:
> 
> ```
> {
>   "id": "12345",
>   "name": "foobar",
>   "description": "hello"
> }
> ```
> 
> and put:
> 
> ```
> {
>   "id": "12345",
>   "name": "foobar"
> }
> ```
> 
> Then description will be null, the condition will evaluate to false and the description will remain as `"hello"`, right? I guess I'm not sure why you need these conditionals at all...

Yes, i got it what you mean, i think you are right here. I've removed it already. Thank you.

mrdpopov (commented):
Please rename oai-pmh-set.raml according to the DTO name

mrdpopov (approved):

dcrossleyau:
@IlliaDaliek Please fix the RAML error that was introduced causing the example to fail.

These error messages will continue to spew on master branch and on subsequent PRs.

PR #122, opened 2020-08-18
MODOAIPMH-210: Implement endpoint for getting values of all required filtering condition types.

IlliaDaliek:
### PURPOSE
Implement the endpoint for getting filtering condition values from inventory storage.
Values are retrieved from next uris: 
- /locations;
- /ill-policies;
- /material-types;
- /instance-types;
- /instance-formats.
https://issues.folio.org/browse/MODOAIPMH-210


ramls/schemas/dto/filteringConditionValue.json
```diff
@@ -0,0 +1,16 @@
...
+  "properties": {
+    "name": {
+      "description": "Filtering condition name",
+      "type": "string"
+    },
+    "values": {
+      "description": "Filtering condition value",
+      "type": "array"
+    }
+  }
```
craigmcnally, line 15: I'm guessing both of these are required fields?

ramls/schemas/dto/filteringConditionValueCollection.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Filtering condition DTO Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "setsFilteringConditions" : {
+      "description": "list of filtering condition values",
+      "type": "array",
+      "id": "filteringConditionValuesList",
```
craigmcnally, line 10: This is a nitpick but the "List" part at the end of the property id feels superfluous to me.

src/main/java/org/folio/oaipmh/dao/PostgresClientFactory.java
```diff
@@ -36,17 +36,15 @@
 
   private static final String DEFAULT_SCHEMA_PROPERTY = "search_path";
 
-  private static final int POOL_SIZE = 5;
+  private static final int POOL_SIZE = 20;
```
craigmcnally, line 39: Deja vu... is this same change in multiple PRs?

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
```diff
@@ -42,4 +75,76 @@ public SetServiceImpl(final SetDao setDao) {
...
+    String tenant = okapiHeaders.get(OKAPI_TENANT);
+    String token = okapiHeaders.get(OKAPI_TOKEN);
+
+    HttpClientRequest httpClientRequest = httpClient.getAbs(okapiUrl.concat(requestUri))
+      .putHeader(OKAPI_TOKEN, token)
+      .putHeader(OKAPI_TENANT, tenant)
+      .putHeader(ACCEPT, APPLICATION_JSON);
+
+    httpClientRequest.handler(response -> {
+      if (response.statusCode() == 200)
```
craigmcnally, line 126: This is personal preference - I'm not sure if there's a project-wide standard, but I prefer to always use brackets for conditionals, even in cases like this.

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
```diff
@@ -1,25 +1,58 @@
...
+  private static final String RESOURCE_TYPES_URI = "/instance-types";
+  private static final String INSTANCE_FORMATS_URI = "/instance-formats";
+
+  private static final String LOCATION_JSON_FIELD_PATH = "locations";
+  private static final String ILL_POLICIES_JSON_FIELD_PATH = "illPolicies";
+  private static final String MATERIAL_TYPES_JSON_FIELD_PATH = "mtypes";
+  private static final String INSTANCE_TYPES_JSON_FIELD_PATH = "instanceTypes";
+  private static final String INSTANCE_FORMATS_JSON_FIELD_PATH = "instanceFormats";
+
+  private static final String LOCATION = "location";
```
craigmcnally, line 44: I don't understand why these need to be restricted to private... it means you need to redefine them again in your test(s).  IMO it's better to open them up to protected, or even public if needed and then use static imports in your test(s)

src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
```diff
@@ -1,30 +1,56 @@
...
-class SetServiceImplTest extends AbstractServiceTest {
+class SetServiceImplTest {
 
+  private static final String TEST_TENANT_ID = "oaiTest";
   private static final String EXISTENT_SET_ID = "16287799-d37a-49fb-ac8c-09e9e9fcbd4d";
   private static final String NONEXISTENT_SET_ID = "a3bd69dd-d50b-4aa6-accb-c1f9abaada55";
   private static final String TEST_USER_ID = "30fde4be-2d1a-4546-8d6c-b468caca2720";
 
+  private static final String LOCATION = "location";
+  private static final String ILL_POLICIES = "illPolicy";
```
craigmcnally, line 49: See my comments above about having to define these more than once.

src/test/java/org/folio/rest/impl/OaiPmhSetImplTest.java
```diff
@@ -71,58 +72,47 @@
     .withSetSpec("update SetSpec");
 
   private final Header tenantHeader = new Header("X-Okapi-Tenant", OAI_TEST_TENANT);
-  private final Header okapiUrlHeader = new Header("X-Okapi-Url", "http://localhost:" + okapiPort);
+  private final Header okapiUrlHeader = new Header("X-Okapi-Url", "http://localhost:" + mockPort);
```
craigmcnally, line 75: can probably make these static too

craigmcnally: or, since you're using `@TestInstance(PER_CLASS)` i *think* you could also just remove static from all these members.

src/test/java/org/folio/rest/impl/OkapiMockServer.java
```diff
@@ -91,6 +92,18 @@
 
   public static final String ERROR_TENANT = "error";
 
+  private static final String LOCATION_URI = "/locations";
+  private static final String ILL_POLICIES_URI = "/ill-policies";
+  private static final String MATERIAL_TYPES_URI = "/material-types";
```
craigmcnally, line 97: See my earlier comment about having to define these constants multiple times because they're private

src/main/java/org/folio/oaipmh/dao/PostgresClientFactory.java
IlliaDaliek: > Deja vu... is this same change in multiple PRs?
Cause i need these changes but PR still not into master, therefore I have to put it into each branch

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
IlliaDaliek: > I don't understand why these need to be restricted to private... it means you need to redefine them again in your test(s). IMO it's better to open them up to protected, or even public if needed and then use static imports in your test(s)
Strongly agree with you, I will fix it now

ramls/schemas/dto/filteringConditionValue.json
IlliaDaliek: No, this never will be taken this from UI or any else via post request for example, this model is used for response only and i use model instead of manual json composing cause vertx JsonObject honestly isn't good into composing json structure correctly, I can show you the particular example if you wish) So, there are no necessity to validate its fields for values presence.

ramls/schemas/dto/filteringConditionValueCollection.json
IlliaDaliek: Okay, i'll rename it)

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
mrdpopov: Well this creates a dependency, unwanted from my perspective, to the code from tests. My preference would be either we  move this to Constancts or use plain strings in tests...

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
IlliaDaliek: Thank you for advice, i've added them here.

src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
IlliaDaliek: I've moved it to Constants.java

src/test/java/org/folio/rest/impl/OaiPmhSetImplTest.java
IlliaDaliek: Thank you, corrected

src/test/java/org/folio/rest/impl/OkapiMockServer.java
IlliaDaliek: Thank you, corrected

ramls/schemas/dto/filteringConditionValue.json
craigmcnally: OK, that's fine.  I'm not saying you have to, but consider adding it though as I don't think it hurts anything and it might be nice to be able to look at the schema and know that both properties will always be present.

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
craigmcnally: These are unit tests aren't they?  Isn't there an inherent dependency (even if via the indirection of API) on the implementation from the test anyway?

Regardless, the point is, repeating the same thing over and over again is a PITA to maintain... if you want to change it, you need to change it in N places.

Moving these into a Constants class is an acceptable compromise.

ramls/schemas/dto/filteringConditionValue.json
IlliaDaliek: > OK, that's fine. I'm not saying you have to, but consider adding it though as I don't think it hurts anything and it might be nice to be able to look at the schema and know that both properties will always be present.
Okay, i will add this before merge, thank you

mrdpopov (approved):

PR #123, opened 2020-08-19
MODOAIPMH-70: Fix incorrect number of records returned by listrecords…

mrdpopov:
## Purpose

Fix incorrect number of records returned by listrecords with resumption token for marc21

Jira: https://issues.folio.org/browse/MODOAIPMH-216


craigmcnally (approved):

MikhailFokanov (changes requested):
Please decline this PR, as @IlliaDaliek fixed this bug it in jira 216 and will merge it within b3.0 merging.

PR #126, opened 2020-08-27
MODOAIPMH-201: Enrich sets endpoints with filtering conditions entites.

IlliaDaliek:
### PURPOSE
Enrich sets endpoint with filtering condition entites -> allow to save set with filtering conditions array and retrieve with it as well.
https://issues.folio.org/browse/MODOAIPMH-201


src/main/resources/liquibase/tenant/scripts/2020-08-20--12-00-enhance-set-table.xml
```diff
@@ -0,0 +1,40 @@
...
+               schemaName="${database.defaultSchemaName}"
+               tableName="filtering_condition"/>
+  </changeSet>
+
+  <changeSet id="2020-08-20-add-filtering-conditions-column-into-set-table"
+             author="Illia Daliek">
+    <addColumn schemaName="${database.defaultSchemaName}"
+               tableName="set">
+      <column name="filtering_conditions"
+              type="text"/>
```
IlliaDaliek, line 19: Here the text data type is used instead of  json cause the jooq-vertx-classic-reactive library has an issue with generating sql query. When json or jsonb data types are used there is a necessety of using data types binding and convertes and here the problems take place. Actually at last library release (5.1.2) it says that jooq can process JSON and JSONB types directly but in real it dosn't. So, there is no way for processing such data types wihout using special bindings, but   there is a problem with redundant brace. Actually i'm gonna report this issue to jooq developers and hope we will be able to use json and jsonb columns whithout any chalanges soon.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
```diff
@@ -99,18 +105,19 @@ public SetDaoImpl(final PostgresClientFactory postgresClientFactory) {
...
-    return getQueryExecutor(tenantId).transaction(queryExecutor -> queryExecutor.executeAny(dslContext -> dslContext.insertInto(SET)
-      .set(toDatabaseSetRecord(entry))
-      .onConflict(SET.ID)
-      .doNothing()
-      .returning())
+    return getQueryExecutor(tenantId).transaction(queryExecutor -> queryExecutor
+      .executeAny(dslContext -> dslContext.insertInto(SET_LB)
+        .set(toDatabaseSetRecord(entry))
+        .onConflict(SET_LB.ID)
+        .doNothing()
```
craigmcnally, line 112: I'm not familiar with this client... if there's a conflict I see it's not overwriting but doing nothing, which makes sense.  I'm wondering what the caller would see... Do they get an error response or would it just fail silently?

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
```diff
@@ -143,17 +150,23 @@ private void prepareSetMetadata(FolioSet entry, String userId, InsertType insert
     entry.setUpdatedByUserId(userId);
   }
 
-  private SetRecord toDatabaseSetRecord(FolioSet set) {
-    SetRecord dbRecord = new SetRecord();
+  private SetLbRecord toDatabaseSetRecord(FolioSet set) {
+    SetLbRecord dbRecord = new SetLbRecord();
     dbRecord.setDescription(set.getDescription());
     if (isNotEmpty(set.getId())) {
```
craigmcnally, line 156: is the set's id not required?  This seems odd to me.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
```diff
@@ -47,8 +52,8 @@ public SetDaoImpl(final PostgresClientFactory postgresClientFactory) {
   @Override
   public Future<FolioSet> getSetById(String id, String tenantId) {
     return getQueryExecutor(tenantId).transaction(txQE -> {
-      Condition condition = SET.ID.eq(UUID.fromString(id));
-      return txQE.findOneRow(dslContext -> dslContext.selectFrom(SET)
+      Condition condition = SET_LB.ID.eq(UUID.fromString(id));
```
craigmcnally, line 55: dumb question... what does the LB stand for?

src/test/java/org/folio/oaipmh/common/AbstractSetTest.java
```diff
@@ -0,0 +1,77 @@
...
+    .withValue("fk updated value")
+    .withSetSpec("fk updated setSpec");
+
+  private static List<FilteringCondition> mockFilteringConditions = ImmutableList.of(MOCK_FILTERING_CONDITION);
+  private static List<FilteringCondition> mockUpdatedFilteringConditions = ImmutableList.of(MOCK_UPDATED_FILTERING_CONDITION);
+
+  protected static FolioSet INITIAL_TEST_SET_ENTRY = new FolioSet().withId(EXISTENT_SET_ID)
+    .withName("test name")
+    .withDescription("test description")
+    .withSetSpec("test setSpec")
```
craigmcnally, line 34: In the schema the description for this field suggests that it actually contains a numeric value... Is that always the case?  If so, why is it a string?  Finally, should the test use realistic data?

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: "onConflict(SET_LB.ID).doNothing" actually is redundant here cause validation on already existed id happens before invoking saveSetItem method. I'm going to remove it now.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: This decision was taken because there were conflicts with naming of dto's. It clashes with Set java framework class name and such things make bring the confusions to code. As well the generated by jooq pojo's were renamed to due the same reason. Another point is that it allows to declare variables without declaring the full class path name like org.folio.rest.jaxrs.model.Set set = new ...

src/test/java/org/folio/oaipmh/common/AbstractSetTest.java
IlliaDaliek: It's just a unit tests, there is no matter what content the models have. As well model contains text data, not numbers.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
craigmcnally: OK that makes sense; but what does it stand for?  Would something like OAI_SET make more sense?

src/test/java/org/folio/oaipmh/common/AbstractSetTest.java
craigmcnally: > Generation from the last record with the same matchedId incremented by 1. Starts from 0.
In that case, this description is confusing since it suggests that the value is an integer > 0.

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
IlliaDaliek: LB - means l-liquid b-base: liquibase database entities. The same naming is used in SRS and i think this is okay. I don't think that it is needed to be renamed again.

src/test/java/org/folio/oaipmh/common/AbstractSetTest.java
IlliaDaliek: image
Where do you see Integer?I can't get what you say. Could you please explain in more detail what you mean please?

src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
craigmcnally: ah, ok got it.

src/test/java/org/folio/oaipmh/common/AbstractSetTest.java
craigmcnally: I'm taking about the description in the schema....
https://github.com/folio-org/mod-oai-pmh/blob/2bf24297a079494bd7bee483a8a01e2ec2aae73d/ramls/schemas/dto/folioSet.json#L21

craigmcnally: if that's not accurate please fix the description.

craigmcnally: "Starts at 0", "incremented by 1" suggests to me that it's an integer... 0, 1, 2, 3, ... N

IlliaDaliek: @craigmcnally, it has been fixed.
image

PR #127, opened 2020-08-28
MODOAIPMH-210: Implement endpoint for getting values of all required filtering condition types.

IlliaDaliek:
### PURPOSE
Fix problem with returning only 10 items of each filtering-condition type (locations, resource-types and etc.)
Jira: https://issues.folio.org/browse/MODOAIPMH-210

### DIFF
--- a/src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
+++ b/src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
@@ -4,6 +4,7 @@ import static javax.ws.rs.core.HttpHeaders.ACCEPT;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.folio.oaipmh.Constants.ILL_POLICIES;
 import static org.folio.oaipmh.Constants.ILL_POLICIES_URI;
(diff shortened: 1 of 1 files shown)


dcrossleyau:
@IlliaDaliek Please fix the RAML error that was introduced causing the example to fail.
(Also see [pull/121](https://github.com/folio-org/mod-oai-pmh/pull/121#issuecomment-676830053))

Otherwise these error messages will continue to spew on master branch and on subsequent PRs.
This also causes difficulty for the tools that keep the API docs configuration up-to-date.

craigmcnally:
HttpClient.getAbs(...) is deprecated and will be going away in vertx4.  see https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClient.html#getAbs-java.lang.String-io.vertx.core.Handler-

Consider switching to webClient (probably not part of this PR - just mentioning it here)

IlliaDaliek:
> HttpClient.getAbs(...) is deprecated and will be going away in vertx4. see https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClient.html#getAbs-java.lang.String-io.vertx.core.Handler-
> 
> Consider switching to webClient (probably not part of this PR - just mentioning it here)

@craigmcnally, I appreciate it, i readed about this.

PR #128, opened 2020-08-31
MODOAIPMH-220: Perform uniqueness validation for sets endpoints

IlliaDaliek:
### PURPOSE
Add unqiue constraints for both name and setSpec columns. 
Jira: https://issues.folio.org/browse/MODOAIPMH-220
 **name unique validation**
name unique
 **setSpec unique validation**
set spec unique

### DIFF
--- a/src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
+++ b/src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
@@ -1,7 +1,7 @@
 package org.folio.oaipmh.service.impl;
 
-import static java.util.Objects.nonNull;
 import static java.lang.String.format;
+import static java.util.Objects.nonNull;
 import static org.folio.oaipmh.Constants.ILL_POLICIES;
 import static org.folio.oaipmh.Constants.INSTANCE_FORMATS;
 import static org.folio.oaipmh.Constants.INSTANCE_TYPES;
--- a/src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
+++ b/src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
@@ -1,6 +1,7 @@
 package org.folio.oaipmh.service.impl;
 
 import static java.util.Objects.nonNull;
+import static java.lang.String.format;
 import static org.folio.oaipmh.Constants.ILL_POLICIES;
 import static org.folio.oaipmh.Constants.INSTANCE_FORMATS;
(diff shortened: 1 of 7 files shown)


mrdpopov:
1. Can you fix raml erors?
2. Can the error message be something like: %s already exists.

mrdpopov (approved):

src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
```diff
@@ -54,9 +57,12 @@
   private static final String NONEXISTENT_SET_ID = "a3bd69dd-d50b-4aa6-accb-c1f9abaada55";
   private static final String TEST_USER_ID = "30fde4be-2d1a-4546-8d6c-b468caca2720";
 
-  private static final String EXPECTED_NOT_FOUND_MSG = String.format("Set with id '%s' was not found", NONEXISTENT_SET_ID);
-  private static final String EXPECTED_ITEM_WITH_ID_ALREADY_EXISTS_MSG = String.format("Set with id '%s' already exists",
+  private static final String EXPECTED_NOT_FOUND_MSG = format("Set with id '%s' was not found", NONEXISTENT_SET_ID);
+  private static final String EXPECTED_ITEM_WITH_ID_ALREADY_EXISTS_MSG = format("Set with id '%s' already exists",
       EXISTENT_SET_ID);
+  private static final String DUPLICATE_KEY_VALUE_VIOLATES_UNIQUE_CONSTRAINT_ERROR_MSG = "duplicate key value violates unique constraint \"%s\"";
```
craigmcnally, line 63: It would be nice if this was defined in one place and reused instead of repeating the definition

IlliaDaliek: @craigmcnally, in fartther PR requests i've created a separate class for mock data especially for Set functionality testing and after all request will be approved then i will merge all of them i correct order, so such simple issues will be disappear caused the above mentioned reason.

craigmcnally:
It looks like the API tests (rest assured tests) essentially make the service tests redundant,  Is there value in having both?  Just seems like more code to maintain

IlliaDaliek:
> It looks like the API tests (rest assured tests) essentially make the service tests redundant, Is there value in having both? Just seems like more code to maintain

@craigmcnally, yes, they both have the different purpose. The same RestAssured test for example mostly aimed on verifying the correct response status code, content type, returned JSON body and other such things. So, the Service test checks the dao layer work + service work by itself and they are required for SonarCloud coverage (no less then 80%) and especially verifies the models data and exceptions handling behavior.

PR #129, opened 2020-08-31
MODOAIPMH-206: POST /oai-pmh/set returns 500 when empty string is passed

IlliaDaliek:
### PURPOSE
FIx issue with 500 error code responding instead of 422 with proper error message when try to save or update set with empty name or setSpec value.
Jira: https://issues.folio.org/browse/MODOAIPMH-206
image

### DIFF
--- a/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
+++ b/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
@@ -162,7 +162,8 @@ public class OaiPmhSetImpl implements OaiPmhSets, OaiPmhFilteringConditions {
     if(isEmpty(folioSet.getName())) {
       String message = format(ERROR_MSG_TEMPLATE, "name");
       errorsList.add(createError("name", "null", message));
-    } if (isEmpty(folioSet.getSetSpec())) {
+    }
+    if (isEmpty(folioSet.getSetSpec())) {
       String message = format(ERROR_MSG_TEMPLATE, "setSpec");
       errorsList.add(createError("setSpec", "null", message));
     }
--- a/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
+++ b/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
@@ -1,12 +1,21 @@
 package org.folio.rest.impl;
 
+import static java.lang.String.format;
+import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
```diff
@@ -25,6 +34,8 @@
 
   private static final Logger logger = LoggerFactory.getLogger(OaiPmhSetImpl.class);
 
+  private static final String ERROR_MSG_TEMPLATE = "null value in column \"%s\" violates not-null constraint";
```
craigmcnally, line 37: Can we revisit the wording of this?  The client doesn't care how this is stored (e.g. column, jsonb, etc.).  It might be best to mimic RMB here, even if not exact - having similar wording would probably make sense.  Here's an example error generated by RMB for a missing required field:

```
{
  "errors": [
    {
      "message": "must not be null",
      "type": "1",
      "code": "-1",
      "parameters": [
        {
          "key": "primaryServicePoint",
          "value": "null"
        }
      ]
    }
  ]
}
```

src/test/java/org/folio/rest/impl/OaiPmhSetImplTest.java
```diff
@@ -57,6 +59,8 @@
   private static final String EXISTENT_SET_ID = "16287799-d37a-49fb-ac8c-09e9e9fcbd4d";
   private static final String NONEXISTENT_SET_ID = "a3bd69dd-d50b-4aa6-accb-c1f9abaada55";
 
+  private static final String NULL_VALUE_ERROR_MSG_TEMPLATE = "null value in column \"%s\" violates not-null constraint";
```
craigmcnally, line 62: Put this in a constants class so we don't have to repeat it?

IlliaDaliek: I've replied on it into https://github.com/folio-org/mod-oai-pmh/pull/128

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
IlliaDaliek: Got it, i am agree with you, i'll correct it now.

IlliaDaliek: Could you approve this please? It would be more convenient for me to make sensible for user error message after merging other PR's.

PR #135, opened 2020-09-04
MODOAIPMH-220: Perform uniqueness validation for sets endpoints

IlliaDaliek:
### PURPOSE
Modify unique constraints to consider the string case as well: qwe = QWE = QwE - only one of these values can be assigned to field under unique constraint.
Plus small fix regarding totalRecords field into set list response. Now totalRecords shows the records count into the whole system instead of the contained into response only.

### DIFF
--- a/src/main/resources/liquibase/tenant/scripts/2020-09-03--12-00-add-case-insensitive-unique-constraints.xml
+++ b/src/main/resources/liquibase/tenant/scripts/2020-09-03--12-00-add-case-insensitive-unique-constraints.xml
@@ -1,0 +1,29 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.9.xsd">
+
+  <changeSet  id="2020-08-28--12-30-drop-liquibase-unique-constraint-from-both-set_spec-and-name-columns" author="Illia Daliek">
+    <dropUniqueConstraint tableName="set_lb"
+                          schemaName="${database.defaultSchemaName}"
+                          constraintName="set_spec_unique_constraint"
+                          uniqueColumns="set_spec"/>
+    <dropUniqueConstraint tableName="set_lb"
+                          schemaName="${database.defaultSchemaName}"
+                          constraintName="name_unique_constraint"
+                          uniqueColumns="name"/>
+  </changeSet>
+
+  <changeSet  id="2020-08-28--11-00-add-case-insensitive-unique-constraint-to-both-set_spec-and-name-columns" author="Illia Daliek">
+    <sql>
+      CREATE UNIQUE INDEX set_spec_unique_constraint
+      ON ${database.defaultSchemaName}.set_lb (lower(set_spec));
+    </sql>
+    <sql>
+      CREATE UNIQUE INDEX name_unique_constraint
+      ON ${database.defaultSchemaName}.set_lb (lower(name));
+    </sql>
+  </changeSet>
+
+</databaseChangeLog>
--- a/src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
+++ b/src/main/java/org/folio/oaipmh/dao/impl/SetDaoImpl.java
@@ -121,19 +121,21 @@ public class SetDaoImpl implements SetDao {
 
   @Override
   public Future<FolioSetCollection> getSetList(int offset, int limit, String tenantId) {
-    return getQueryExecutor(tenantId).transaction(queryExecutor -> queryExecutor.query(dslContext -> dslContext.selectFrom(SET_LB)
-      .orderBy(SET_LB.UPDATED_DATE.desc())
-      .offset(offset)
-      .limit(limit))
-      .map(this::queryResultToSetCollection));
+    return getQueryExecutor(tenantId).transaction(queryExecutor -> queryExecutor.query(dslContext -> dslContext.selectCount().from(SET_LB))).compose(recordsCount ->
+        getQueryExecutor(tenantId).transaction(queryExecutor -> queryExecutor.query(dslContext -> dslContext.selectFrom(SET_LB)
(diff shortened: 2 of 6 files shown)


craigmcnally:
I'm sure there was good reason to go with liquidbase instead of using the exsiting RMB tooling, but can you fill me in on those reaons?  I ask because it feels a little like you're re-inventing the wheel here a bit - previously we discussed things like managing record metadata, now handling case sensitivity, standard record collections (totalRecord counts), etc.  

I know other teams have also gone this route - is there an opportunity to create a shared library, (e.g. lb-utils) or something to help with some of this?  I imagine these are the sorts of things that others adopting LB will also need to deal with (or already have dealt with).

IlliaDaliek:
> I'm sure there was good reason to go with liquidbase instead of using the exsiting RMB tooling, but can you fill me in on those reaons? I ask because it feels a little like you're re-inventing the wheel here a bit - previously we discussed things like managing record metadata, now handling case sensitivity, standard record collections (totalRecord counts), etc.
> 
> I know other teams have also gone this route - is there an opportunity to create a shared library, (e.g. lb-utils) or something to help with some of this? I imagine these are the sorts of things that others adopting LB will also need to deal with (or already have dealt with).

Thank you for question. Let me explain it. So, there were decision to use library which allows to communicate with database using java objects. Great tools would be the hibernate or Spring JPA, but for known reasons we cannot use these spring related techonologies. So, we have taken the jooq like in SRS for next reasons - it's more convinient to use java classes & "java databse entites" for building quires and prcoessing them as well plus it allows to use less code (for example you have not to declare string queries constants or smth like that). So you just write convinient methods using arrow functions. That is all what relates to jooq. But there was the next problem - jooq needs the database for generating dto, pojo, table entities and so on. But we cannot have access to databse before TenantAPI invoking and the solution was to allow jooq see our future database to generate required data via liquibase which is populating the temporary databse during build. As well liquibase allow us to have database versioning which can be very useful at someday.

craigmcnally:
> Actually there already exists the folio liquibase library - https://github.com/folio-org/folio-liquibase-util

Thanks.  Doesn't look like there's much there yet.  As you add required functionality to this module, please consider the possibility that what you're doing could be useful to others.  If so, generalizing it and adding to these utils would be a good idea.

craigmcnally:
> > I'm sure there was good reason to go with liquidbase instead of using the exsiting RMB tooling, but can you fill me in on those reaons? I ask because it feels a little like you're re-inventing the wheel here a bit - previously we discussed things like managing record metadata, now handling case sensitivity, standard record collections (totalRecord counts), etc.
> > I know other teams have also gone this route - is there an opportunity to create a shared library, (e.g. lb-utils) or something to help with some of this? I imagine these are the sorts of things that others adopting LB will also need to deal with (or already have dealt with).
> 
> @craigmcnally, thank you for question, let me explain it. So, there was the decision to use library which allows to communicate with database using java objects. Great tools would be the hibernate or Spring JPA, but for known reasons we cannot use these spring related techonologies. So, we have taken the jooq like in SRS for next reasons - it's more convinient to use java classes & "java databse entites" for building quires and prcoessing them as well plus it allows to use less code (for example you have not to declare string queries constants or smth like that). So you just write convinient methods using arrow functions. That is all what relates to jooq. But there was the next problem - jooq needs the database for generating dto, pojo, table entities and so on. But we cannot have access to databse before TenantAPI invoking and the solution was to allow jooq see our future database to generate required data via liquibase which is populating the temporary databse during build. As well liquibase allow us to have database versioning which can be very useful at someday.
> 
> Actually there already exists the folio liquibase library - https://github.com/folio-org/folio-liquibase-util

At this point it's probably too late to really turn back, but to me it sounds like you're jumping through a lot of hoops and accepting several of compromises for this one feature.  I wonder if it's really worth it...  The time spent on adapting and writing tooling to work with LB/jooq is time that's not spent on actual feature development.  It also means that the rest of the community that might not be familiar with these technologies has a bit of a learning curve if they ever have to work in this code (or even participate in code reviews).

src/main/resources/liquibase/tenant/scripts/2020-09-03--12-00-add-case-insensitive-unique-constraints.xml
```diff
@@ -0,0 +1,29 @@
...
+  </changeSet>
+
+  <changeSet  id="2020-08-28--11-00-add-case-insensitive-unique-constraint-to-both-set_spec-and-name-columns" author="Illia Daliek">
+    <sql>
+      CREATE UNIQUE INDEX set_spec_unique_constraint
+      ON ${database.defaultSchemaName}.set_lb (lower(set_spec));
+    </sql>
+    <sql>
+      CREATE UNIQUE INDEX name_unique_constraint
+      ON ${database.defaultSchemaName}.set_lb (lower(name));
```
craigmcnally, line 25: is `lower(...)` good enough here?  What about accents or other diacritical marks?

e.g. `facade` and `Facade` would not be unique, but `façade` would be.

mrdpopov: It looks like you are right. There is, however,  a critical need for this functionality to be merged for the upcoming demo. I suggest that we merge it now and we'll create a ticket to investigate this point further.

mrdpopov (approved):

src/main/resources/liquibase/tenant/scripts/2020-09-03--12-00-add-case-insensitive-unique-constraints.xml
mrdpopov: Quote from PO:
Our PO doesn’t insist on handling such cases and even says that there are a lot of words in French, German etc, where the same word with and without (or other) diacritical mark, umlaut and so on can have another meaning (e.g. tâche ‘task’ — tache ‘spot’), so it is arguable, whether it should be considered unique or not, unlike case sensitivity.

craigmcnally (approved):

src/main/resources/liquibase/tenant/scripts/2020-09-03--12-00-add-case-insensitive-unique-constraints.xml
craigmcnally: It's worth noting that this deviates from the vast majority of the project's storage modules.  I wasn't part of the decision making process and can't speak to the rationale behind the decision.

PR #152, opened 2020-10-15
MODOAIPMH-245: Edge module doesn't return the data, even if mod-oai-pmh does

IlliaDaliek:


oleksandr-yatsenko1 (approved):
Looks good for me.

PR #158, opened 2020-10-29
MODOAIPMH-251: Release 3.2.2 version mod-oai-pmh

IlliaDaliek:


andrii-novytskyi (approved):

KVupp:
As I mentioned before, it would have been good if these procedures were followed:https://dev.folio.org/guidelines/release-procedures/
- It is usually suggested to increment to the next minor version not bugfix versions
- There is no description in the PR
- Add everyone from the team for the PR review 
- There is no changelog
- The tag is incorrect in the pom.xml

PR #162, opened 2020-11-09
MODOAIPMH-250 Invalid number of identifiers are returned when get ListIdentifiers

IlliaDaliek:
### PURPOSE
Fix issue with incorrect number of items being returned when perform ListIdentifiers request.

_Cannot be covered with unit tests cause number of returned records are mocked. 
Test coverage will be added in Karate integration test project._ 

Jira: https://issues.folio.org/browse/MODOAIPMH-250

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
@@ -77,7 +77,7 @@ public class GetOaiIdentifiersHelper extends AbstractHelper {
         updatedBefore,
         null,
         request.getOffset(),
-        batchSize,
+        batchSize+1,
         response -> response.bodyHandler(bh -> {
           try {
             final OAIPMH oaipmh = buildListIdentifiers(request, bh.toJsonObject());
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
```diff
@@ -77,7 +77,7 @@
         updatedBefore,
         null,
         request.getOffset(),
-        batchSize,
+        batchSize+1,
```
andrii-novytskyi, line 80: do we able to write a unit test for this?

IlliaDaliek: > do we able to write a unit test for this?

In PR header i've described why it cannot be covereged with unit tests.

Igor-Gorchakov (approved):

src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
KVupp: Just a nitpick but add a comment as to why it is being done, and probably better to do the addition  of 1 where the variable is being declared for more readability

KVupp: > unit tests cause number of returned records are mocked.
If you are mocking them using the vert.x mock server, it can still be handled to return the mocked response based on the batch size param

illia-borysenko (approved):

src/main/java/org/folio/oaipmh/helpers/GetOaiIdentifiersHelper.java
IlliaDaliek: It is required to request one more record to know is there any records present yet or to validate next records requesting by comparing the excess batchSize+1 record form prev request with first record id in next request. By the end the last recod is deleted from response, only id of it is saved to resumptionToken. Therefore if we are requesting 10 oai-pmh records per response then module requests the srs(or any each) for 10+1 record end returns only first 10 records and id of 11's record in resumptionToken. Bug was that batchsize was equal to records per response, therefore the 10's record was deleted and 9 instead of 10 records were returned.

KVupp: I meant to add a comment in the code not in the PR, in the future no one will know why we added a +1 :)

PR #163, opened 2020-11-11
MODOAIPMH-252: Mod OAI_PMH container fail after call

IlliaDaliek:
### PURPOSE
Fix issue with memory leaking and add better handling of errors from both SRS and inventory-storage.

Jira: https://issues.folio.org/browse/MODOAIPMH-252


src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -1948,6 +1956,50 @@ void shouldRespondWithBadArgument_whenGetOaiRecordsMarc21WithHoldingsWithInvalid
     System.setProperty(REPOSITORY_MAX_RECORDS_PER_RESPONSE, currentValue);
   }
 
+  @Test
+  void shouldRespondWithInternalServerError_whenGetOaiRecordsMarc21WithHoldingsWithErrorResponseFromInventoryStorage() {
```
IlliaDaliek, line 1960: But here in such response the normal informative message is returned.

MikhailFokanov (approved):
Looks good.

Igor-Gorchakov (approved):

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -514,33 +528,62 @@ private boolean isSkipSuppressed(Request request) {
...
         savePromise.future().toCompletionStage().thenRun(completePromise::complete);
       }
+      databaseWriteStream.invokeDrainHandler();
     });
 
     return completePromise;
   }
 
+  private Object getValueFrom(Object obj, String fieldName) {
+    Field field = requireNonNull(ReflectionUtils.findField(requireNonNull(obj.getClass()), fieldName));
```
IlliaDaliek, line 563: Here the reflection is used by the reason that we need to have an access to vert.x "waiters" objects which are accumulated when batches are saved to database, the handling batches from inventory view is performed match faster versus saving to database. By this reason in some time we got a lot of waiters objects wich holds many of others and this leads to OutOfMemory.

MikhailFokanov: In future we can consider using static AtomicInteger to count the number of current db requests. It will be more readable in code, but less reliable because wouldn't take into account other requests.

ViktorSoroka07: @IlliaDaliek, could you please add some comment in the source code as well.

PR #165, opened 2020-11-23
MODOAIPMH-240: Newest git update introduces build loop

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-240

### PURPOSE
Resolve conflicts between time zones when running tests locally plus fixing by these changes the bug related to records requesting with invalid time borders.

The SRS client converts the date to UTC therefore the end date will be as next: passed date param - time differnce between env system time zone(STZ) and UTC.

### Approach
Before passing the dates to SRC client we add the time diff between current system time zone and UTC and further the same time diff will be subtracted by the SRC client back and by the end the same date which has been passed as param will be used as search criteria for records.
Example how it worked before: 
- STZ is +02:00
- before param which is passed equals to 2018-12-12T00:00:00z
- SRC client under the hood converts the 'before' to UTC, therefore it subtracts 2 hours of time zones difference from 'before' date(cuse UTC is 2 hours less then current time zone)
The 'before' param which is passed by SRC client was 2018-12-11T22:00:00z instead of expected 2018-12-12T00:00:00z

Example how it works now after fix:
- STZ is +02:00
- before param which is passed equals to 2018-12-12T00:00:00z
- the diff between time zones is added to 'before' param, 'before' param now which is passed to SRC client is 2018-12-12T02:00:00z (2 hours more)
- SRC client under the hood converts the 'before' to UTC, therefore it subtracts 2 hours of time zones difference from 'before' date(cuse UTC is 2 hours less then current time zone)
'before' param which is passed by SRC client now is 2018-12-12T00:00:00z which is expected.

The same is applicable for until param as well.

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
@@ -54,8 +54,8 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
       final boolean deletedRecordsSupport = RepositoryConfigurationUtil.isDeletedRecordsEnabled(request);
       final boolean suppressedRecordsSupport = getBooleanProperty(request.getOkapiHeaders(), REPOSITORY_SUPPRESSED_RECORDS_PROCESSING);
 
-      final Date updatedAfter = request.getFrom() == null ? null : convertStringToDate(request.getFrom(), false);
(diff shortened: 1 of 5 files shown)


Igor-Gorchakov (approved):

andrii-novytskyi (approved):

ViktorSoroka07:
@IlliaDaliek, please don't forget to remove brances.

PR #166, opened 2020-11-25
MODOAIPMH-258: Clean data for outdated requests from instance table

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-258

### PURPOSE
Add job for cleaning expiring instances. Expiration period is 1 day.
In scope of this story the refactoring has been done for database communication layer, all DB related methods that had been spreaded among the code has been moved to single InstancesDao and simplified with jooq library. All new introduced code covered with unit tests. As well some improvements have been done for tests classes.

### APPROACH
Add new DB table which stores the request_id (holds all instances related to particular harvester) and last updated date, after instances have been retrieved the last_updated_date is updated and after one day passed the job starts to look for expired request ids and if some of them occures then job performs cleaning of all instances by expired request ids.


src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -160,6 +160,49 @@ public static MarcWithHoldingsRequestHelper getInstance() {
     return promise.future();
   }
 
+  private void updateLastUpdateDateForRequestMetadata(String requestId, PostgresClient postgresClient, String tenantId) {
```
IlliaDaliek, line 163: Need to change with additional jooq dao method which will not take more then 10 lines of code at maximum

src/test/java/org/folio/rest/impl/CleanUpJobTest.java
```diff
@@ -0,0 +1,205 @@
...
+  void shouldReturn204AndClearExpiredInstances(VertxTestContext testContext) {
+    RequestSpecification request = createBaseRequest(CLEAN_UP_INSTANCES_PATH, null);
+    request.when()
+      .post()
+      .then()
+      .statusCode(204);
+  }
+
+  //will add batch save
+  private Future<RequestMetadataLb> saveRequestMetadata(RequestMetadataLb requestMetadata) {
```
IlliaDaliek, line 166: Need to be moved to the dao layer and be reused at MarcWithHoldingsRequestHelper and test classes as well.

src/test/java/org/folio/rest/impl/CleanUpJobTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+  private RequestMetadataLbRecord toDbRecord(RequestMetadataLb pojo) {
+    RequestMetadataLbRecord record = new RequestMetadataLbRecord();
+    record.setId(pojo.getId());
+    record.setRequestId(pojo.getRequestId());
+    record.setLastUpdatedDate(pojo.getLastUpdatedDate());
+    return record;
+  }
+
+  private Future<Boolean> deleteRequestMetadataById(String requestId) {
```
IlliaDaliek, line 181: The same relates to comment above

src/test/java/org/folio/oaipmh/common/AbstractInstancesTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static org.folio.rest.impl.OkapiMockServer.OAI_TEST_TENANT;
+
+public abstract class AbstractInstancesTest {
+
+  protected static final int INSTANCES_EXPIRATION_TIME_IN_SECONDS = 7200;
```
KVupp, line 26: so the request Id which was not called in last 2 hours is deleted from the instances table, what is the behaviour if they request the data with resumption token after 2 hours? How will the API respond?

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -66,8 +66,14 @@
 
   /** Map containing OAI-PMH verb and corresponding helper instance. */
   private static final Map<VerbType, VerbHelper> HELPERS = new EnumMap<>(VerbType.class);
+  private static final int INSTANCES_EXPIRATION_TIME_IN_SECONDS = 7200;
```
KVupp, line 69: so the request Id which was not called in last 2 hours is deleted from the instances table, what is the behavior if they request the data with resumption token after 2 hours? How will the API respond?

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -33,6 +33,7 @@
 
 import javax.ws.rs.core.Response;
 
+import io.vertx.core.*;
```
KVupp, line 36: Don't use the * import

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
KVupp, line 69: What is the API behaviour if the request with the resumption token is invoked after 2 hours?

src/main/resources/liquibase/tenant/scripts/2020-11-19--18-00-create-request-metadata-table.xml
```diff
@@ -0,0 +1,22 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+  <changeSet id="2020-11-19--18-00-create-request-metadata-table.xml" author="Illia Borysenko">
+    <createTable tableName="request_metadata_lb">
+      <column name="request_id" type="uuid">
+        <constraints primaryKey="true" nullable="false"/>
```
KVupp, line 10: should we also have a reference to the instance table with a foreign key?

src/main/java/org/folio/oaipmh/service/impl/InstancesServiceImpl.java
```diff
@@ -0,0 +1,86 @@
...
+      });
+  }
+
+  @Override
+  public Future<RequestMetadataLb> saveRequestMetadata(RequestMetadataLb requestMetadata, String tenantId) {
+    return instancesDao.saveRequestMetadata(requestMetadata, tenantId);
+  }
+
+  @Override
+  public Future<RequestMetadataLb> updateRequestMetadataByRequestId(String requestId, RequestMetadataLb requestMetadataLb,
```
KVupp, line 57: What is the caller of this method?

KVupp:
> and last updated date, after instances have been retrieved the last_updated_date is updated 

What about the failure scenario? What if a job is unable to complete the retrieval or in fact what happens if we fail to update the last_updated_date in the metadata table?

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
KVupp: The description mentions 1 day but here it just has 2 hours?

KVupp:
Also as discussed in the standup today, Please add some notes to the read me section with what is happening with these tables. Thanks!

KVupp:
@IlliaDaliek  Thanks for the details in the description , they are really helpful before going through the code

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -1979,27 +1986,28 @@ void shouldRespondWithInternalServerError_whenGetOaiRecordsMarc21WithHoldingsWit
...
-      .get()
-      .then()
-      .statusCode(500)
-      .extract()
-      .asString();
-    String expectedMessage = format(EXPECTED_ERROR_MESSAGE, "source-record-storage");
-    assertTrue(body.contains(expectedMessage));
-    System.setProperty(REPOSITORY_MAX_RECORDS_PER_RESPONSE, currentValue);
-  }
+//  @Test
```
illia-borysenko, line 1989: Please remove commented code

IlliaDaliek:
> > and last updated date, after instances have been retrieved the last_updated_date is updated
> 
> What about the failure scenario? What if a job is unable to complete the retrieval or in fact what happens if we fail to update the last_updated_date in the metadata table?

So, metadata table is very important here and if some error occurs while saving or updating entities within metadata table then whole oai-pmh request going to be failed. So only when all parts works well then harvester can get the records.

src/main/java/org/folio/oaipmh/service/impl/InstancesServiceImpl.java
IlliaDaliek: MarcWithHoldingsRequestHelper at 158 line. So if we have the first request of marc21_withholdings then we create request metadata entity at opposite case we updating existing one by request id got from resumptionToken

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
IlliaDaliek: For now it will lead to error because we will not be able to update request metadata by request id got from resumptionToken by the reason that such requestId has been deleted already by this moment. To prevent this we should as well validate resumtionToken requestId before running main request processing logic but this should be implemnted in scope of https://issues.folio.org/browse/MODOAIPMH-164.

IlliaDaliek: Thank you for good note. Changed to 86400 seconds. Regarding your first question I've explained it above.

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
IlliaDaliek: I just forgot to setup the code format settings at my laptop) I've used it to work for last days. Thank you for note. Fixed.

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
IlliaDaliek: It has been done just for research purposes, I've uncommented this code. Thank you.

src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -2088,4 +2098,35 @@ private ResumptionTokenType makeResumptionTokenRequestsAndVerifyCount(List<Heade
...
     return resumptionToken;
   }
+
+  @Autowired
+  public void setInstancesService(InstancesService instancesService) {
+    this.instancesService = instancesService;
+  }
+
+
+//  @ParameterizedTest
```
ViktorSoroka07, line 2108: Please remove commented code.

src/main/resources/liquibase/tenant/scripts/2020-11-19--18-00-create-request-metadata-table.xml
IlliaDaliek: Yes, this makes sense, I added a foreign key constraint and change the logic of deleting instances respectively. Thank you.

src/main/java/org/folio/oaipmh/dao/InstancesDao.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Returns request ids which last updated date is 'expirationPeriodInSeconds' seconds less than the current date.
+   *
+   * @param tenantId                  - tenant id
+   * @param expirationPeriodInSeconds - expiration time in seconds
+   * @return list of expired request ids
+   */
+  Future<List<String>> getExpiredRequestIds(String tenantId, int expirationPeriodInSeconds);
+
+  Future<RequestMetadataLb> saveRequestMetadata(RequestMetadataLb requestMetadata, String tenantId);
```
illia-borysenko, line 21: Add documentation for all methods

src/main/resources/liquibase/tenant/scripts/2020-11-19--18-00-create-request-metadata-table.xml
```diff
@@ -0,0 +1,32 @@
...
+        <constraints nullable="false"/>
+      </column>
+    </createTable>
+  </changeSet>
+
+  <changeSet id="2020-11-19--18-00-change-json-to-text-column-in-instances-table" author="Illia Daliek">
+    <modifyDataType tableName="instances" columnName="json" newDataType="text" schemaName="${database.defaultSchemaName}"/>
+  </changeSet>
+
+  <changeSet id="2020-11-19--18-00-change-varchar-to-uuid-column-in-instances-table" author="">
```
Igor-Gorchakov, line 22: Please fill 'author' section

src/main/java/org/folio/oaipmh/service/InstancesService.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Performs cleaning of instances which place at DB more then 'expirationTimeSeconds' seconds.
+   *
+   * @param tenantId - tenant id
+   * @return request ids associated with removed instances
+   */
+  Future<List<String>> cleanExpiredInstances(String tenantId, int expirationTimeSeconds);
+
+  Future<RequestMetadataLb> saveRequestMetadata(RequestMetadataLb requestMetadata, String tenantId);
```
Igor-Gorchakov, line 22: add javadoc

illia-borysenko, line 22: Please add documentation for all methods

src/test/java/org/folio/oaipmh/service/impl/InstancesServiceImplTest.java
```diff
@@ -0,0 +1,235 @@
...
+  void shouldReturnFutureWithEmptyList_whenThereNoExpiredRequestIds(VertxTestContext testContext) {
+    testContext.verify(() -> instancesService.cleanExpiredInstances(TEST_TENANT_ID, ZERO_EXPIRED_INSTANCES_TIME)
+      .onComplete(testContext.succeeding(ids -> {
+        assertTrue(ids.isEmpty());
+        testContext.completeNow();
+      })));
+  }
+
+  @Test
+  void shouldReturnFutureWithExpiredIds_whenThereExpiredRequestIdsArePresented(VertxTestContext testContext) {
```
illia-borysenko, line 96: I think the word "there" is not necessary for this method name

IlliaDaliek: I am agree

illia-borysenko (approved):

src/main/java/org/folio/oaipmh/service/impl/InstancesServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+  public InstancesServiceImpl(InstancesDao instancesDao) {
+    this.instancesDao = instancesDao;
+  }
+
+  @Override
+  public Future<List<String>> cleanExpiredInstances(String tenantId, int expirationTimeSeconds) {
+    Promise<List<String>> promise = Promise.promise();
+    instancesDao.getExpiredRequestIds(tenantId, expirationTimeSeconds)
+      .onSuccess(ids -> {
+        List<Future> futures = new ArrayList<>();
```
ViktorSoroka07, line 37: Sonar complains about code here as a code smell. Please check out.

IlliaDaliek: I've already discussed this with back-end guys, this is not needed o be fixed.

Igor-Gorchakov (approved):

PR #169, opened 2020-12-10
MODOAIPMH-259 HTML encoded entities in records make the OAI-PMH requests crash

Igor-Gorchakov:
**Purpose**
Fix the case when HTML encoded entities in records make the OAI-PMH requests crash

**Solution**
HTML encoded entities do not crash the module, the reason for failures is that the 'updatedDate' & 'createdDate' come in the old date-time format.
I fixed the AbstractStorageHelper that converts dates everywhere in the module. I have written unit tests for the old & new formats of data-time.


IlliaDaliek (approved):

KVupp:
@Igor-Gorchakov  I am a little surprised, According to the ticket(https://issues.folio.org/browse/MODOAIPMH-259) the record is being returned from mod-oai-pmh but is not returned from edge-oai-pmh. Why does edge module returning a 504 relate to date format, can you please explain?

andrii-novytskyi (approved):

Igor-Gorchakov:
> @Igor-Gorchakov I am a little surprised, According to the ticket(https://issues.folio.org/browse/MODOAIPMH-259) the record is being returned from mod-oai-pmh but is not returned from edge-oai-pmh. Why does edge module returning a 504 relate to date format, can you please explain?

Hi @KVupp , according to the MODOAIPMH-259 when the attached SRS record retrieved the mod-oai-pmh gets timeout and the response returned with 504. I wrote a unit test and the module gets timeout as described in Jira. 
The reason for timeout is ParseException that gets thrown while parsing 'createdDate' or 'updatedDate' in the old time format. I just fixed the parser adding one more format, as it was fixed in generate-marc-utils earlier. 
Now both oai&edge return 200ok response:

image
image

KVupp:
> > @Igor-Gorchakov I am a little surprised, According to the ticket(https://issues.folio.org/browse/MODOAIPMH-259) the record is being returned from mod-oai-pmh but is not returned from edge-oai-pmh. Why does edge module returning a 504 relate to date format, can you please explain?
> 
> Hi @KVupp , according to the MODOAIPMH-259 when the attached SRS record retrieved the mod-oai-pmh gets timeout and the response returned with 504. I wrote a unit test and the module gets timeout as described in Jira.
> The reason for timeout is ParseException that gets thrown while parsing 'createdDate' or 'updatedDate' in the old time format. I just fixed the parser adding one more format, as it was fixed in generate-marc-utils earlier.
> Now both oai&edge return 200ok response:
> 
> image
> image

This is what I see in the ticket:
2. Get the record from /oai/records. **The record is returned**, even though not really as expected and then in the next step 
3. Get the record from edge-oai. -> This is when they get 504

IlliaDaliek:
> > > @Igor-Gorchakov I am a little surprised, According to the ticket(https://issues.folio.org/browse/MODOAIPMH-259) the record is being returned from mod-oai-pmh but is not returned from edge-oai-pmh. Why does edge module returning a 504 relate to date format, can you please explain?
> > 
> > 
> > Hi @KVupp , according to the MODOAIPMH-259 when the attached SRS record retrieved the mod-oai-pmh gets timeout and the response returned with 504. I wrote a unit test and the module gets timeout as described in Jira.
> > The reason for timeout is ParseException that gets thrown while parsing 'createdDate' or 'updatedDate' in the old time format. I just fixed the parser adding one more format, as it was fixed in generate-marc-utils earlier.
> > Now both oai&edge return 200ok response:
> 
> This is what I see in the ticket:
> 2. Get the record from /oai/records. **The record is returned**, even though not really as expected and then in the next step
> 3. Get the record from edge-oai. -> This is when they get 504
> 
> I saw your unit test. The difference is there is no special character like in the screen shot. I think the text pasted in the ticket doesn't capture that character.
Hi @KVupp , we are confident in the root cause that Igor described. The timeout problem was related only to the new RMB date format which was not handled in oai-pmh. So, we spent today about 2 hours testing this issue properly. We put the same record and instance attached at the jira ticket and linked them. So, we haven't got the timeout anymore and we got the same record from both edge module and oai-pmh only.

KVupp:
ok Thanks, If you all are confident with the testing, go ahead and merge it.

PR #170, opened 2020-12-15
MODOAIPMH-123: Datestamp in response doesn't correspond to time granularity

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-123
### PURPOSE
Make header datestamps be affected by time granularity setting, i.e. header datestamp should match time granularity format.

### APPROACH
Set date to header type in a particular date format.

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
@@ -1,57 +1,10 @@
 package org.folio.oaipmh.helpers;
 
-import io.vertx.core.Future;
-import io.vertx.core.Promise;
-import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.JsonObject;
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
-import org.apache.commons.lang.time.DateUtils;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.Pair;
-import org.folio.oaipmh.MetadataPrefix;
-import org.folio.oaipmh.Request;
-import org.folio.oaipmh.ResponseConverter;
-import org.folio.oaipmh.helpers.response.ResponseHelper;
-import org.folio.oaipmh.helpers.storage.StorageHelper;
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
```diff
@@ -419,27 +434,20 @@ protected HeaderType addHeader(String identifierPrefix, Request request, JsonObj
...
   protected boolean canResumeRequestSequence(Request request, Integer totalRecords, JsonArray instances) {
     Integer prevTotalRecords = request.getTotalRecords();
     boolean isDeletedRecords = isDeletedRecordsEnabled(request);
     int firstPosition = 0;
     return instances != null && instances.size() > 0
-        && (totalRecords >= prevTotalRecords || StringUtils.equals(request.getNextRecordId(),
-            isDeletedRecords ? storageHelper.getId(instances.getJsonObject(firstPosition))
-                : storageHelper.getRecordId(instances.getJsonObject(firstPosition))));
+      && (totalRecords >= prevTotalRecords || StringUtils.equals(request.getNextRecordId(),
+      isDeletedRecords ? storageHelper.getId(instances.getJsonObject(firstPosition))
```
andrii-novytskyi, line 449: usage of deprecated storageHelper.getId method?

andrii-novytskyi:
maddening to watch changed imports in PR, we should configure the style as soon as possible

andrii-novytskyi (approved):

src/main/java/org/folio/oaipmh/helpers/AbstractHelper.java
IlliaDaliek: I don't think that this method is deprecated. It would be better to remove the @Deprecated annotation from such method. It is used and works well. I'll remove the annotation.

src/main/java/org/folio/oaipmh/helpers/storage/StorageHelper.java
```diff
@@ -80,7 +80,6 @@ static StorageHelper getInstance() {
 
   JsonArray getRecordsItems(JsonObject entries);
 
-  @Deprecated
```
andrii-novytskyi, line 83: what is the reason why it was deprecated?

Igor-Gorchakov (approved):

PR #172, opened 2020-12-21
MODOAIPMH-266 Upgrade to RMB 32

illia-borysenko:
https://issues.folio.org/browse/MODOAIPMH-266

## Purpose
Upgrade to the latest RMB to make use of the vert.x 4
## Approach
Update codebase according to changes documented in update notes for version 32 https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-320
## TODO
Temporary disable some unit test, should be enabled again in scope of this PR


descriptors/ModuleDescriptor-template.json
```diff
@@ -64,7 +64,7 @@
     },
     {
       "id" : "_tenant",
-      "version" : "1.2",
+      "version" : "2.0",
```
KVupp, line 67: There is also a suggested changed to make the tenantAPI asynchronous. Please check https://github.com/folio-org/mod-configuration/pull/102/files for more details. It would be good to get it done

pom.xml
```diff
@@ -115,8 +115,8 @@
     <sonar.coverage.exclusions>**/BatchStreamWrapper.java</sonar.coverage.exclusions>
     <!-- Plugin versions -->
     <aspectj.version>1.9.6</aspectj.version>
-    <raml-module-builder.version>31.1.5</raml-module-builder.version>
-    <vertx.version>3.9.4</vertx.version>
+    <raml-module-builder.version>32.0.0-SNAPSHOT</raml-module-builder.version>
+    <vertx.version>4.0.0</vertx.version>
```
KVupp, line 119: This file also needs the below change:
```RMB-728 In pom.xml update aspectj-maven-plugin <configuration> with <complianceLevel>11</complianceLevel>.```

KVupp (commented):
There are no changes relating to the below bullet point from the upgrade notes:
```OKAPI-943 When calling .any, .all or .join replace io.vertx.core.CompositeFuture by org.folio.okapi.common.GenericCompositeFuture, replace raw type by actual type and remove @SuppressWarnings("rawtypes").```

InstancesServiceImpl.java ,SetServiceImpl,ModTenantAPI use the compositeFuture.all

pom.xml
illia-borysenko: It's already was updated to version 11

descriptors/ModuleDescriptor-template.json
KVupp: This is not a mandatory change for this release. So if this is the only change holding up the release, let's not wait. We can incorporate it later

IlliaDaliek (approved):

IlliaDaliek:
Looks good

PR #180, opened 2021-01-08
MODOAIPMH-278: Revert RMB 32.0.0 due to RMB bugs.

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-278

### PURPOSE
Initial-load request cannot be processed after RMB update since there is a bug that breaks the connection between inventory-storage and oai-pmh after some time. A small number of instances can be processed normally, but when a request started to take more than half of a minute approximately the problems comes up.

### APPROACH
Revert the commits which relate to RMB update.


src/test/java/org/folio/rest/impl/OkapiMockServer.java
```diff
@@ -299,8 +304,6 @@ private void handleRecordStorageResultPostResponse(RoutingContext ctx) {
       successResponse(ctx, getJsonObjectFromFile(SOURCE_STORAGE_RESULT_URI + TWO_RECORDS_ONE_CANNOT_BE_CONVERTED_TO_XML_JSON));
     } else if (instanceIds.contains(DEFAULT_INSTANCE_ID)) {
       successResponse(ctx, getJsonObjectFromFile(SOURCE_STORAGE_RESULT_URI + SRS_RECORD));
-    } else if (instanceIds.contains(INSTANCE_ID_NO_SRS_RECORDS)) {
```
andrii-novytskyi, line 302: does this code relates to rmb revert?

PR #183, opened 2021-01-18
MODOAIPMH-282: On 7M records first response to the initial requests with marc21_withholdings metadataPrefix takes more than 20 min

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-282

### PURPOSE 
Fix error due to great count of logs.


MikhailFokanov (approved):
Looks good.

PR #184, opened 2021-01-25
MODOAIPMH-282, MODOAIPMH-284: speed up request processing & add the possibility to return response when the first batch has been downloaded

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-282
Jira: https://issues.folio.org/browse/MODOAIPMH-284

### PURPOSE
Fix problem with long marc21_withholdings request processing.  Return the response when the first batch will be downloaded with letting downloading instances be continued in the background. As well these changes include the reverting of MODOAIPMH-273.


src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -267,13 +266,6 @@ void setUpBeforeEach() {
     RestAssured.config().decoderConfig(DecoderConfig.decoderConfig());
   }
 
-  @AfterEach
-  void cleanUp(VertxTestContext testContext) {
```
IlliaDaliek, line 271: Since all instances are linked via requestId which is unique in the scope of each single test case then the data conflict is impossible here, therefore the method should be deleted.

MikhailFokanov (approved):

src/main/resources/liquibase/tenant/scripts/2020-11-19--18-00-create-request-metadata-table.xml
```diff
@@ -40,4 +40,18 @@
                              referencedColumnNames="request_id" onDelete="CASCADE" onUpdate="NO ACTION"/>
   </changeSet>
 
+  <changeSet id="2021-01-19--14-00-add-boolean-field-to-request-metadata" author="Illia Daliek">
+    <addColumn tableName="request_metadata_lb">
+      <column name="stream_ended" type="boolean" defaultValueBoolean="true">
```
andrii-novytskyi, line 45: Is default value should be true for defaultValueBoolean?

andrii-novytskyi (approved):

src/main/resources/liquibase/tenant/scripts/2020-11-19--18-00-create-request-metadata-table.xml
IlliaDaliek: Yes

illia-borysenko-ebsco (approved):

PR #185, opened 2021-01-28
Fix DB pool conflicts

IlliaDaliek:
### PURPOSE
Tests that are run after OaiPmhImplTest and uses the database cannot obtain normal connections cause some threads after OaiPmhImplTest still hand the connections from the pool associated with the test tenant oaiTest.

### PURPOSE
Reset the connection pool before each test that uses the database.

### DIFF
--- a/src/main/java/org/folio/oaipmh/dao/PostgresClientFactory.java
+++ b/src/main/java/org/folio/oaipmh/dao/PostgresClientFactory.java
@@ -36,6 +36,8 @@ public class PostgresClientFactory {
 
   private static final int POOL_SIZE = 20;
 
+  private static boolean shouldResetPool = false;
+
   private static final Map<String, PgPool> POOL_CACHE = new HashMap<>();
 
   private Vertx vertx;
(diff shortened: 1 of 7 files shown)


illia-borysenko (approved):

src/main/java/org/folio/oaipmh/dao/PostgresClientFactory.java
```diff
@@ -36,6 +36,8 @@
 
   private static final int POOL_SIZE = 20;
 
+  private static boolean shouldResetPool = false;
```
andrii-novytskyi, line 39: please add ToDo comment to remove this field in the future, with a description of why it present here,  as this field needed only for tests

src/test/java/org/folio/oaipmh/service/impl/SetServiceImplTest.java
```diff
@@ -62,12 +58,16 @@
   private static final int mockPort = NetworkUtils.nextFreePort();
 
   private static PostgresClientFactory postgresClientFactory;
-
-  private SetDao setDao;
   private SetService setService;
```
Igor-Gorchakov, line 61: private SetService setService = new new SetServiceImpl(new SetDaoImpl(postgresClientFactory));

PR #187, opened 2021-01-28
MODOAIPMH-281: Release mod-oai-pmh 3.3.0

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-281

### PURPOSE
Merge changes after maven release command.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,12 @@
+## 3.3.0 (Released) 2021-01-28
+
+This release includes the bug fixes and improvements regarding marc21_withholdings metadata prefix processing.
+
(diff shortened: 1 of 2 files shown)


Igor-Gorchakov (approved):

illia-borysenko-ebsco (approved):

IlliaDaliek:
Merging with 3 hotspot issues. Sonar says that the password specified in the embedded Postgres maven plugin should not be hardcoded since any user can get it from our open source project but this password doesn't matter and is used only for DB during maven build in order to create a pojo's and etc in pair with liquibase plugin. Such an error hasn't appeared previously, so I will find out how to make sonar ignore these lines in pom.xml at my further pull request.

PR #188, opened 2021-02-08
MODOAIPMH-285: Update RMB version

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-285
Jira: https://issues.folio.org/browse/MODOAIPMH-289

### **PURPOSE**
Update module to RMB 32.0.2
As well in the scope of the story, the problems with the WebClient closing were fixed.

### **FYI** 
Since all related changes in this PR were restored they all were already viewed(see https://github.com/folio-org/mod-oai-pmh/pull/172). The only MarcWithHoldingsRequestHelper and BatchStreamWrapper files have been changed and should be viewed


Igor-Gorchakov (approved):

MikhailFokanov (approved):

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
```diff
@@ -82,15 +85,16 @@ public SetServiceImpl(final SetDao setDao) {
...
-    Future<JsonObject> instanceFormatsFuture = getFilteringConditionValues(INSTANCE_FORMATS_URI, vertxHttpClient, okapiHeaders);
-
-    CompositeFuture.all(locationFuture, illPoliciesFuture, materialTypesFuture, resourceTypeFuture, instanceFormatsFuture)
+    WebClient webClient = WebClient.create(Vertx.vertx());
+    Future<JsonObject> locationFuture = getFilteringConditionValues(LOCATION_URI, webClient, okapiHeaders);
+    Future<JsonObject> illPoliciesFuture = getFilteringConditionValues(ILL_POLICIES_URI, webClient, okapiHeaders);
+    Future<JsonObject> materialTypesFuture = getFilteringConditionValues(MATERIAL_TYPES_URI, webClient, okapiHeaders);
+    Future<JsonObject> resourceTypeFuture = getFilteringConditionValues(RESOURCE_TYPES_URI, webClient, okapiHeaders);
+    Future<JsonObject> instanceFormatsFuture = getFilteringConditionValues(INSTANCE_FORMATS_URI, webClient, okapiHeaders);
+
```
craigmcnally, line 94: Another place where WebClient is not closed.

craigmcnally (changes requested):
Still leaking web cilents

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
IlliaDaliek: Very good point. Thank you, Craig!

craigmcnally (commented):
I also see that in src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java you're still using HttpClient, and on line 339 you don't close the client.  Please switch to webClient in this class for consistency and close the client

craigmcnally:
There are 5 code smells. Please address these too.  The logging related ones should be quick and easy.  The code duplication one might take some shuffling around of code but I can't imagine it being much work either.

src/main/java/org/folio/oaipmh/service/impl/SetServiceImpl.java
MikhailFokanov: @craigmcnally, I believe it is closed some lines below. Or I missed something?

  List<Future<JsonObject>> futures = Arrays.asList(locationFuture, illPoliciesFuture, materialTypesFuture,
      resourceTypeFuture, instanceFormatsFuture);
    GenericCompositeFuture.all(futures)
      .onComplete(result -> {
        if (result.failed()) {
          webClient.close();
          promise.fail(result.cause());
        } else {
          List<SetsFilteringCondition> values = new ArrayList<>();
          values.add(jsonObjectToSetsFilteringCondition(locationFuture.result(), LOCATION_JSON_FIELD_PATH, LOCATION));
          values.add(jsonObjectToSetsFilteringCondition(illPoliciesFuture.result(), ILL_POLICIES_JSON_FIELD_PATH, ILL_POLICIES));
          values
            .add(jsonObjectToSetsFilteringCondition(materialTypesFuture.result(), MATERIAL_TYPES_JSON_FIELD_PATH, MATERIAL_TYPES));
          values
            .add(jsonObjectToSetsFilteringCondition(resourceTypeFuture.result(), INSTANCE_TYPES_JSON_FIELD_PATH, INSTANCE_TYPES));
          values.add(jsonObjectToSetsFilteringCondition(instanceFormatsFuture.result(), INSTANCE_FORMATS_JSON_FIELD_PATH,
              INSTANCE_FORMATS));

          FilteringConditionValueCollection filteringConditionValueCollection = new FilteringConditionValueCollection()
            .withSetsFilteringConditions(values);
          webClient.close();
          promise.complete(filteringConditionValueCollection);
        }
      });

IlliaDaliek:
> There are 5 code smells. Please address these too. The logging related ones should be quick and easy. The code duplication one might take some shuffling around of code but I can't imagine it being much work either.

I don't think that it makes sense to move to a separate private method two lines with closing client and batchstream.

IlliaDaliek:
> I also see that in src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java you're still using HttpClient, and on line 339 you don't close the client. Please switch to WebClient in this class for consistency and close the client
We cannot use WebClient with our current implementation of WriteStream. That's why we use HttpClient instead.

MikhailFokanov:
> I also see that in src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java you're still using HttpClient, and on line 339 you don't close the client. Please switch to webClient in this class for consistency and close the client

We cannot get rid of HttpClient here beacuse it is not a typical REST call, but "batching" (or "streaming") call. We can reiceve 7 mln instances during this one call and save them to the database by batches. WebClient cannot do it in such way. This is for what Vertx HttpClient is intended for and it is not deprecated for this use case.

MikhailFokanov:
> > There are 5 code smells. Please address these too. The logging related ones should be quick and easy. The code duplication one might take some shuffling around of code but I can't imagine it being much work either.
> 
> I don't think that it makes sense to move to a separate private method two lines with closing client and batchstream.

Why not, if sonar thinks that it is bad? We should reconfigure sonar, or fix what it suggests

IlliaDaliek:
> > > There are 5 code smells. Please address these too. The logging related ones should be quick and easy. The code duplication one might take some shuffling around of code but I can't imagine it being much work either.
> > 
> > 
> > I don't think that it makes sense to move to a separate private method two lines with closing client and batchstream.
> 
> Why not, if sonar thinks that it is bad? We should reconfigure sonar, or fix what it suggests

I prefer the first suggestion. Previously the Sonar didn't tell anything and now it started to consider such code as a code smell. The same relates to logging. It is so weird.

IlliaDaliek:
> > > > There are 5 code smells. Please address these too. The logging related ones should be quick and easy. The code duplication one might take some shuffling around of code but I can't imagine it being much work either.
> > > 
> > > 
> > > I don't think that it makes sense to move to a separate private method two lines with closing client and batchstream.
> > 
> > 
> > Why not, if sonar thinks that it is bad? We should reconfigure sonar, or fix what it suggests
> 
> I prefer the first suggestion. Previously the Sonar didn't tell anything and now it started to consider such code as a code smell. The same relates to logging. It is so weird. That is why I didn't fix them.

craigmcnally:
Are you saying you're not going to make the fixes to the code smells?

IlliaDaliek:
> Are you saying you're not going to make the fixes to the code smells?
No, I will do of course. I mean that I don't understand why Sonar started to consider some code as code smell but previously it didn't.

IlliaDaliek:
> > Are you saying you're not going to make the fixes to the code smells?
> 
> No, I will do of course. I mean that I don't understand why Sonar started to consider some code as code smell but previously it didn't.

craigmcnally:
> > Are you saying you're not going to make the fixes to the code smells?
> 
> No, I will do of course. I mean that I don't understand why Sonar started to consider some code as code smell but previously it didn't.

I think with the logging ones it has to do with the move to log4j2

IlliaDaliek:
> > > Are you saying you're not going to make the fixes to the code smells?
> > 
> > 
> > No, I will do of course. I mean that I don't understand why Sonar started to consider some code as code smell but previously it didn't.
> 
> I think with the logging ones it has to do with the move to log4j2

Yes, it makes sense, I guess the same.

craigmcnally (approved):

craigmcnally:
technically the coverage is a bit low, but it's not a huge deal given the overall coverage will still be quite good.  In the interest of moving this along I'm fine with letting it slide.

IlliaDaliek:
> technically the coverage is a bit low, but it's not a huge deal given the overall coverage will still be quite good. In the interest of moving this along I'm fine with letting it slide.

Regarding code coverage, it is because of the switching to the new RMB approach like replacing HttpClient with WebClient, logger's changes and etc. All this stuff brought changes to our code and sonar considers it as a new code. But the business logic stayed the same and it was covered well. Sonar started to consider some classes again from the scratch. But before this change, the same classes weren't pointed with less than 80% code coverage. For example, we can take the MarcXmlMapper class. In git diff, we see that only logger-related stuff was changed, but if we will back to sonar code coverage analytic we will see that it started to check not only logger-related stuff but and all other class code as well and this is not correct. We can see that it swears on "catch" and "finally" blocks. But when this class was made and merged these changes were verified at that moment already and changes had more than 80% code coverage. In any further PR's these changes weren't considered for code coverage as it wasn't the new code already. Same situation with some other classes listed in the current sonar code coverage list. If we put all these things aside the real code coverage will be high.

PR #190, opened 2021-02-12
MODOAIPMH-291: Revert RMB 32 update

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-291

### PURPOSE
Since we encountered the same problem with JsonParser exceptions as we did previously it is needed to revert the RMB.


MikhailFokanov (approved):
Looks good.

PR #192, opened 2021-02-23
MODOAIPMH-292: Incorrect cursor in resumptionToken

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-292

### PURPOSE
Fix incorrect cursor value returned within resumption token. 
Add an index to the request_id column in order to improve the request processing with date range parameters. Previously without index, the request responded with the timeout error.

The autoincremented id field was added to the instances table and is used for ordering instead of using the insatnce_id column when selecting instance ids from the table. Performing ordering by insatnce_id is unstable by the reason that while we requesting batches the new instances are saved in the background and the latest saved ids may come up at the top of the SELECT query result which brokes the behavior of using resumption token. In simple words, the stable order using the insatnce_id column for the "ORDER BY" statement can be achieved only when we have all instances already saved. Therefore while we ordering them by instance_id at the time when they are saved the order is changed as well. The autoincremented id fixed this and guarantee that the order will not be broken while the new incoming instances are saved.

As well since the SRS may respond with 502 when we request 500 records, the re-requesting of srs calls were implemented. In total there are 50 attempts to re-request SRS and only when all 50 attempts are failed the oai-pmh request processing is going to be canceled.

### DIFF
--- a/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
+++ b/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
@@ -1,5 +1,21 @@
 package org.folio.oaipmh.processors;
 
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+import static javax.ws.rs.core.HttpHeaders.ACCEPT;
+import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static org.folio.oaipmh.Constants.NEXT_RECORD_ID_PARAM;
+import static org.folio.oaipmh.Constants.OFFSET_PARAM;
+import static org.folio.oaipmh.Constants.OKAPI_TENANT;
+import static org.folio.oaipmh.Constants.OKAPI_TOKEN;
+import static org.folio.oaipmh.Constants.REPOSITORY_MAX_RECORDS_PER_RESPONSE;
+import static org.folio.oaipmh.Constants.REPOSITORY_SUPPRESSED_RECORDS_PROCESSING;
+import static org.folio.oaipmh.Constants.REQUEST_ID_PARAM;
+import static org.folio.oaipmh.Constants.UNTIL_PARAM;
+import static org.folio.oaipmh.helpers.RepositoryConfigurationUtil.getBooleanProperty;
+
 import java.lang.reflect.Field;
 import java.math.BigInteger;
 import java.time.OffsetDateTime;
@@ -8,20 +24,20 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import javax.ws.rs.core.Response;
 
-import io.vertx.core.WorkerExecutor;
 import org.apache.commons.collections4.CollectionUtils;
 import org.folio.oaipmh.Request;
+import org.folio.oaipmh.client.SourceStorageSourceRecordsClient;
 import org.folio.oaipmh.dao.PostgresClientFactory;
 import org.folio.oaipmh.helpers.AbstractHelper;
 import org.folio.oaipmh.helpers.RepositoryConfigurationUtil;
@@ -29,7 +45,6 @@ import org.folio.oaipmh.helpers.records.RecordMetadataManager;
 import org.folio.oaipmh.helpers.response.ResponseHelper;
 import org.folio.oaipmh.helpers.streaming.BatchStreamWrapper;
 import org.folio.oaipmh.service.InstancesService;
-import org.folio.oaipmh.client.SourceStorageSourceRecordsClient;
 import org.folio.rest.jooq.tables.pojos.Instances;
 import org.folio.rest.jooq.tables.pojos.RequestMetadataLb;
 import org.folio.rest.persist.PostgresClient;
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -162,7 +163,7 @@ private MarcWithHoldingsRequestHelper() {
       }
 
       updateRequestMetadataFuture.onSuccess(res -> {
-        boolean isFirstBatch = resumptionToken == null || request.getRequestId() == null;
```
IlliaDaliek, line 165: The second condition is unnecessary here. The first one tells the same. Request object can contain both at once or no one of them.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -200,7 +201,7 @@ private void processBatch(Request request, Context context, Promise<Response> oa
           return;
         }
 
-        if (!instances.isEmpty() && instances.get(0).getString(INSTANCE_ID_FIELD_NAME).equals(request.getNextRecordId())) {
+        if (!firstBatch && (CollectionUtils.isNotEmpty(instances) && !instances.get(0).getString(INSTANCE_ID_FIELD_NAME).equals(request.getNextRecordId()))) {
```
IlliaDaliek, line 204: We should respond with an error in the case when instance ids are NOT equal.

craigmcnally:
It would be great if you could add an "Approach" section to the PR description, with summary list of what's being done here... 

e.g.
* Add an ID column to the instance_ids table for purposes of sorting / maintaining the same order on the initial and subsequent requests.  This is simple integer incrementing from 0, 1, 2, ... N.
* Fix the stale resumption token check such that .... etc.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -113,11 +115,11 @@
   public static final MarcWithHoldingsRequestHelper INSTANCE = new MarcWithHoldingsRequestHelper();
   private final Vertx vertx;
 
-  public static final int POLLING_TIME_INTERVAL = 500;
-
-  public static final int MAX_WAIT_UNTIL_TIMEOUT = 1000*60*20;
+  private static final int REQUEST_ATTEMPTS = 50;
```
craigmcnally, line 118: Can you explain why these were changed to private?

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
craigmcnally: Consider adding a comment in the code instead of here in the PR.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -219,12 +222,8 @@ private void processBatch(Request request, Context context, Promise<Response> oa
         final SourceStorageSourceRecordsClient srsClient = new SourceStorageSourceRecordsClient(request.getOkapiUrl(),
           request.getTenant(), request.getOkapiToken());
 
-        Future<Map<String, JsonObject>> srsResponse = Future.future();
-        if (CollectionUtils.isNotEmpty(instances)) {
-          srsResponse = requestSRSByIdentifiers(srsClient, instancesWithoutLast, deletedRecordSupport);
-        } else {
-          srsResponse.complete();
-        }
+        Future<Map<String, JsonObject>> srsResponse = requestSRSByIdentifiers(srsClient, instancesWithoutLast, deletedRecordSupport);
```
craigmcnally, line 225: probably don't need this `srsResponse` object at all, just call `onSuccess(...)` directly on the future returned by `requestSRSByIdentifiers(...)`

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -677,45 +674,71 @@ private HttpClientRequest createEnrichInventoryClientRequest(HttpClient httpClie
...
-              .filter(Objects::nonNull)
-              .map(JsonObject.class::cast)
-              .forEach(jo -> result.put(jo.getJsonObject("externalIdsHolder")
-                .getString(INSTANCE_ID_FIELD_NAME), jo));
+      srsClient.postSourceStorageSourceRecords("INSTANCE", deletedRecordSupport, listOfIds, srsResponse -> {
+        int statusCode = srsResponse.statusCode();
+        String statusMessage = srsResponse.statusMessage();
+        if (statusCode != 200) {
+          if(statusCode >= 500) {
+            String warnMessage = "Got error response form SRS, status code: " + statusCode + ", status message: {}" + statusMessage;
```
craigmcnally, line 690:
```suggestion
            String warnMessage = "Got error response form SRS, status code: " + statusCode + ", status message: " + statusMessage;
```

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -677,45 +674,71 @@ private HttpClientRequest createEnrichInventoryClientRequest(HttpClient httpClie
...
+            logger.warn(warnMessage);
+            if (attemptsCount.get() <= 0) {
+              String errorMessage = "SRS didn't respond with expected status code after " + REQUEST_ATTEMPTS + " attempts. Canceling further request processing.";
+              srsClient.close();
+              handleException(promise, new IllegalStateException(errorMessage));
+              return;
+            }
+            logger.warn("Trying to request SRS again, attempts left: {}", attemptsCount.decrementAndGet());
+            doPostRequestToSrs(srsClient, deletedRecordSupport, listOfIds, attemptsCount, promise);
```
craigmcnally, line 699: We probably want to have some retry backoff instead of just immediately trying again.  Ideally it would be configurable, but to reduce scope creep we can use a static constant.  It should be at least a second or two IMO

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: It is simple class fields, they should be private. My guess here is that these fields were created among the code and then they were moved to constant by IDEA hotkeys with selected default access as public.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: Yes, you are right. Thank you for your note.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: Thank you for the suggestion. I added vertx timer here with a 1-second delay.

craigmcnally (approved):

PR #195, opened 2021-03-05
MODOAIPMH-300: Not all instances ids are saved for initial harvest

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-300
Jira: https://issues.folio.org/browse/MODOAIPMH-301

### PURPOSE
Fix issue with the hanging of srs response handler which caused the harvest to be canceled. 
The root cause is that when srs responds with a large response body it may hang for an unknown reason and the body is not written to the response, therefore we can get the body and invoke body handler which causes the code hanging and canceling harvesting as result.

### APPROCH

Set the idle timeout of srs request to 20 seconds (instead of 5000) in order to invoke the exception handler (by causing timeout exception) when there a problem with receiving response body from SRS. Make a retry of srs call in the exception handler.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,12 +76,12 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-sql-client</artifactId>
-        <version>${vertx.version}-FOLIO</version>
+        <version>3.9.4-FOLIO</version>
       </dependency>
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-pg-client</artifactId>
-        <version>${vertx.version}-FOLIO</version>
+        <version>3.9.4-FOLIO</version>
       </dependency>
       <dependency>
         <groupId>org.junit</groupId>
@@ -119,7 +119,7 @@
     <!-- Plugin versions -->
     <aspectj.version>1.9.6</aspectj.version>
     <raml-module-builder.version>31.1.5</raml-module-builder.version>
-    <vertx.version>3.9.4</vertx.version>
+    <vertx.version>3.9.6</vertx.version>
     <vertx-completable-future.version>0.1.2</vertx-completable-future.version>
     <apache-httpclient.version>4.5.11</apache-httpclient.version>
     <apache-commons-lang3.version>3.9</apache-commons-lang3.version>
--- a/src/main/java/org/folio/oaipmh/Constants.java
+++ b/src/main/java/org/folio/oaipmh/Constants.java
@@ -23,7 +23,7 @@ public final class Constants {
 
   public static final String REPOSITORY_BASE_URL = "repository.baseURL";
   public static final String REPOSITORY_MAX_RECORDS_PER_RESPONSE = "repository.maxRecordsPerResponse";
-  public static final String REPOSITORY_HTTP_REQUEST_RETRY_ATTEMPTS = "repository.httpRequestRetryAttempts";
+  public static final String REPOSITORY_SRS_HTTP_REQUEST_RETRY_ATTEMPTS = "repository.srsHttpRequestRetryAttempts";
   public static final String REPOSITORY_NAME = "repository.name";
   public static final String REPOSITORY_ADMIN_EMAILS = "repository.adminEmails";
   public static final String REPOSITORY_TIME_GRANULARITY = "repository.timeGranularity";
--- a/src/main/java/org/folio/oaipmh/Constants.java
+++ b/src/main/java/org/folio/oaipmh/Constants.java
@@ -8,8 +8,6 @@ public final class Constants {
     throw new IllegalStateException("This class holds constants only");
   }
 
-
-
   /**
    * Strict ISO Date and Time with UTC offset.
    * Represents {@linkplain org.openarchives.oai._2.GranularityType#YYYY_MM_DD_THH_MM_SS_Z YYYY_MM_DD_THH_MM_SS_Z} granularity
@@ -25,6 +23,7 @@ public final class Constants {
 
   public static final String REPOSITORY_BASE_URL = "repository.baseURL";
   public static final String REPOSITORY_MAX_RECORDS_PER_RESPONSE = "repository.maxRecordsPerResponse";
+  public static final String REPOSITORY_HTTP_REQUEST_RETRY_ATTEMPTS = "repository.httpRequestRetryAttempts";
   public static final String REPOSITORY_NAME = "repository.name";
   public static final String REPOSITORY_ADMIN_EMAILS = "repository.adminEmails";
   public static final String REPOSITORY_TIME_GRANULARITY = "repository.timeGranularity";
--- a/src/main/java/org/folio/oaipmh/Constants.java
+++ b/src/main/java/org/folio/oaipmh/Constants.java
@@ -8,6 +8,7 @@ public final class Constants {
     throw new IllegalStateException("This class holds constants only");
   }
 
+
   /**
    * Strict ISO Date and Time with UTC offset.
    * Represents {@linkplain org.openarchives.oai._2.GranularityType#YYYY_MM_DD_THH_MM_SS_Z YYYY_MM_DD_THH_MM_SS_Z} granularity
@@ -21,9 +22,12 @@ public final class Constants {
   public static final String OKAPI_TENANT = "X-Okapi-Tenant";
   public static final String OKAPI_TOKEN = "X-Okapi-Token";
 
+  public static final String JAXB_MARSHALLER_FORMATTED_OUTPUT = "jaxb.marshaller.formattedOutput";
+  public static final String JAXB_MARSHALLER_ENABLE_VALIDATION = "jaxb.marshaller.enableValidation";
   public static final String REPOSITORY_BASE_URL = "repository.baseURL";
   public static final String REPOSITORY_MAX_RECORDS_PER_RESPONSE = "repository.maxRecordsPerResponse";
   public static final String REPOSITORY_SRS_HTTP_REQUEST_RETRY_ATTEMPTS = "repository.srsHttpRequestRetryAttempts";
+  public static final String REPOSITORY_SRS_CLIENT_IDLE_TIMEOUT_SEC = "repository.srsClientIdleTimeoutSec";
   public static final String REPOSITORY_NAME = "repository.name";
   public static final String REPOSITORY_ADMIN_EMAILS = "repository.adminEmails";
   public static final String REPOSITORY_TIME_GRANULARITY = "repository.timeGranularity";
--- a/src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
+++ b/src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
@@ -19,16 +19,21 @@ import io.vertx.core.logging.LoggerFactory;
 //TODO Should be replaced with the SourceStorageSourceRecordsClient from mod-source-record-storage-client
 // when it's upgraded to RMD version 32 and Vert.x version 4
 public class SourceStorageSourceRecordsClient {
+
+  private static final Logger logger = LoggerFactory.getLogger(SourceStorageSourceRecordsClient.class);
+
   private static final String GLOBAL_PATH = "/source-storage/source-records";
+  private static final int DEFAULT_SRS_TIMEOUT = 10000;
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MS = 2000;
+  private static final int DEFAULT_IDLE_TIMEOUT_SEC = 20;
+
   protected String tenantId;
   protected String token;
   protected String okapiUrl;
   private HttpClientOptions options;
   private HttpClient httpClient;
 
-  private static final Logger logger = LoggerFactory.getLogger(SourceStorageSourceRecordsClient.class);
-
-  public SourceStorageSourceRecordsClient(String okapiUrl, String tenantId, String token, boolean keepAlive, int connTO, int idleTO) {
+  public SourceStorageSourceRecordsClient(String okapiUrl, String tenantId, String token, boolean keepAlive, int connTO, int idleTimeoutSec) {
     this.tenantId = tenantId;
     this.token = token;
     this.okapiUrl = okapiUrl;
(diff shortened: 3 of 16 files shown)


src/test/java/org/folio/rest/impl/OaiPmhImplTest.java
```diff
@@ -2196,9 +2199,10 @@ void shouldRespondWithInternalServerError_whenGetOaiRecordsMarc21WithHoldingsWit
       .statusCode(500)
       .extract()
       .asString();
-    String expectedMessage = format(EXPECTED_ERROR_MESSAGE, "source-record-storage");
+    String expectedMessage = "SRS didn't respond with expected status code after 1 attempts. Canceling further request processing.";
```
IlliaDaliek, line 2202: Unit test for re-requesting srs was written before this PR
It is called shouldRerequestSrsWhenItRespondedWith500_whenGetOaiRecordsMarc21WithHoldingsWithErrorResponseAndThenSuccessResponseFromSrs

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -2,38 +2,32 @@
...
-  private String okapiUrl;
+  protected String tenantId;
+  protected String token;
+  protected String okapiUrl;
   private HttpClientOptions options;
   private HttpClient httpClient;
 
+  private static final Logger logger = LoggerFactory.getLogger(SourceStorageSourceRecordsClient.class);
+
   public SourceStorageSourceRecordsClient(String okapiUrl, String tenantId, String token, boolean keepAlive, int connTO, int idleTO) {
```
craigmcnally, line 31: Since units were a problem here, it may be a good idea to incorporate the units into the variable names... e.g. `idleTimeoutSec`, `connTimeoutMs`

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -69,8 +67,14 @@ public void postSourceStorageSourceRecords(String idType, Boolean deleted, List
       buffer.appendString(ClientHelpers.pojo2json(List));
     }
 
+
```
craigmcnally, line 70:
```suggestion
```

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -441,6 +439,7 @@ private void getNextBatch(String requestId, Request request, boolean firstBatch,
       return completePromise.future();
     }
 
+
```
craigmcnally, line 442:
```suggestion
```

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -47,10 +41,14 @@ public SourceStorageSourceRecordsClient(String okapiUrl, String tenantId, String
   }
 
   public SourceStorageSourceRecordsClient(String okapiUrl, String tenantId, String token) {
-    this(okapiUrl, tenantId, token, true, 2000, 5000);
+    this(okapiUrl, tenantId, token, true, 2000, 20);
```
craigmcnally, line 44: Instead of hardcoding these timeout values, can we make them constants?

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -69,8 +67,14 @@ public void postSourceStorageSourceRecords(String idType, Boolean deleted, List
       buffer.appendString(ClientHelpers.pojo2json(List));
     }
 
+
     HttpClientRequest request = this.httpClient.postAbs(this.okapiUrl + "/source-storage/source-records" + queryParams.toString());
     request.handler(responseHandler);
+    request.exceptionHandler(e-> {
+      logger.error("SRS response error 1234: " + e.getMessage(), e);
```
craigmcnally, line 74: what is the significance of error 1234?

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -69,8 +67,14 @@ public void postSourceStorageSourceRecords(String idType, Boolean deleted, List
...
+      exceptionHandler.handle(e);
+    });
+    request.setTimeout(10000);
```
craigmcnally, line 77: See comment above about using constants instead of integer literals

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -694,74 +692,82 @@ private HttpClientRequest createEnrichInventoryClientRequest(HttpClient httpClie
...
+          String errorMsg = getErrorFromStorageMessage("source-record-storage", srsResponse.request()
+            .absoluteURI(), srsResponse.statusMessage());
+          srsClient.close();
+          handleException(promise, new IllegalStateException(errorMsg));
+          return;
         }
-        srsResponse.bodyHandler(getSrsResponseBodyHandler(srsClient, promise));
-      });
+        srsResponse.bodyHandler(buffer -> handleSrsResponse(srsClient, promise, buffer));
+      }, e->  retrySRSRequest(srsClient, vertx, deletedRecordSupport, listOfIds, attemptsCount, retryAttempts, promise, 400, "Error in SRS response"));
```
craigmcnally, line 725: I think it's misleading to always report 400 status here...   It's probably better to not report a status than to potentially report the wrong one.

craigmcnally, line 725: If I read this right, the exception handler provided here will only be invoked upon request exceptions (`request.exceptionHandler(...)`).  In this case there is no status code.  Furthermore, the message provided `Error in SRS response` is inaccurate.  `Error making SRS request` would be more accurate.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -694,74 +692,82 @@ private HttpClientRequest createEnrichInventoryClientRequest(HttpClient httpClie
...
-      } catch (DecodeException ex) {
-        String msg = "Invalid json has been returned from SRS, cannot parse response to json.";
-        handleException(promise, new IllegalStateException(msg, ex));
-      } catch (Exception e) {
-        handleException(promise, e);
-      } finally {
-        srsClient.close();
+  private void retrySRSRequest(SourceStorageSourceRecordsClient srsClient, Vertx vertx, boolean deletedRecordSupport, List<String> listOfIds, AtomicInteger attemptsCount, int retryAttempts, Promise<Map<String, JsonObject>> promise, int statusCode, String statusMessage) {
+    String warnMessage = "Got error response form SRS, status code: " + statusCode + ", status message: " + statusMessage;
+    logger.debug(warnMessage);
```
craigmcnally, line 733: It's probably better to log this before calling this method.  The way this currently stands, in some cases you might log a message at ERROR level, then a different (also inaccurate - see earlier comments) message at WARN level.

src/main/java/org/folio/oaipmh/Constants.java
```diff
@@ -25,6 +23,7 @@ private Constants() {
 
   public static final String REPOSITORY_BASE_URL = "repository.baseURL";
   public static final String REPOSITORY_MAX_RECORDS_PER_RESPONSE = "repository.maxRecordsPerResponse";
+  public static final String REPOSITORY_HTTP_REQUEST_RETRY_ATTEMPTS = "repository.httpRequestRetryAttempts";
```
craigmcnally, line 26: `httpRequestRetryAttempts` is a little misleading - The module makes several HTTP requests.  This configuration only applies to the call to SRS.

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
craigmcnally:
```suggestion
      logger.error("Exception making SRS request: " + e.getMessage(), e);
```

pom.xml
```diff
@@ -76,12 +76,12 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-sql-client</artifactId>
-        <version>${vertx.version}-FOLIO</version>
+        <version>3.9.4-FOLIO</version>
```
craigmcnally, line 79: I didn't see this mentioned in the PR description... is this a necessary part of this PR?  If so, can you please explain why?

IlliaDaliek: When I had a discussion with Adam and Julian they suggested to use the latest vertx version with major version 3 (3.9.6).
But dependencies with the FOLIO prefix for 3.9.6 don't exist. So we changed the vertx version property value to be 3.9.6 and left some dependencies with 3.9.4-FOLIO.

src/main/java/org/folio/oaipmh/Constants.java
IlliaDaliek: Okay, it makes sense. I added the "srs" prefix to the property name.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: Agree, corrected.

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
IlliaDaliek: Done

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
IlliaDaliek: Thanks for the note, initially this log was added just for testing of implemented logic work. I deleted it since we have loggin in the "srs retry method".

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
IlliaDaliek: Yes, done.

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
IlliaDaliek: Agree, corrected.

MikhailFokanov (approved):
Looks good.

illia-borysenko-ebsco (approved):

pom.xml
craigmcnally: In the future, please add this context to the PR description so that it's clear why the change is being made.  Without this information the PR reviewer is left wondering if the change was intentional, or a mistake.

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -2,55 +2,58 @@
...
 // when it's upgraded to RMD version 32 and Vert.x version 4
 public class SourceStorageSourceRecordsClient {
+
+  private static final Logger logger = LoggerFactory.getLogger(SourceStorageSourceRecordsClient.class);
+
   private static final String GLOBAL_PATH = "/source-storage/source-records";
-  private String tenantId;
-  private String token;
-  private String okapiUrl;
+  private static final int DEFAULT_SRS_TIMEOUT = 10000;
```
craigmcnally, line 26: why no units on this one?

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
```diff
@@ -69,8 +72,10 @@ public void postSourceStorageSourceRecords(String idType, Boolean deleted, List
       buffer.appendString(ClientHelpers.pojo2json(List));
     }
 
-    HttpClientRequest request = this.httpClient.postAbs(this.okapiUrl + "/source-storage/source-records" + queryParams.toString());
+    HttpClientRequest request = this.httpClient.postAbs(this.okapiUrl + GLOBAL_PATH + queryParams.toString());
```
craigmcnally, line 75: I realize this was already there and you're just using it but `GLOBAL_PATH` is a lousy name for this.  Something like `SOURCE_RECORDS_PATH` is IMO much more appropriate

craigmcnally (dismissed):
A couple minor comments related to variable names.  Approving now, but please take my comments into consideration.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
craigmcnally: Can we remove this extra newline?

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
craigmcnally: I still see you passing a hard-coded 400 into this method in the case where there is no response code - because the error occurred while writing the request.  Changing the log level from WARN to DEBUG is not sufficient to fix this IMO.

craigmcnally (changes requested):
Please address the issues raised about the logging in retry method

craigmcnally:
@IlliaDaliek I noticed that you've introduce new changes here after the PR was approved.  Please re-request reviews from those that previously approved.  This feels a little like scope creep here.  Strictly speaking this probably should have happened in a separate PR, probably with its own JIRA.  Given that the changes are already here, we might as well go with it.

IlliaDaliek:
> @IlliaDaliek I noticed that you've introduce new changes here after the PR was approved. Please re-request reviews from those that previously approved. This feels a little like scope creep here. Strictly speaking this probably should have happened in a separate PR, probably with its own JIRA. Given that the changes are already here, we might as well go with it.

@craigmcnally, there no serious changes, I moved srs idle timeout to module settings as Mikhail asked me to do that. As well when I verified the "schema update testing" ticket I noticed incorrect logs and fixed them as well in the current PR. And among the code the laterals were used instead of already defined constants, I corrected this too.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: > I still see you passing a hard-coded 400 into this method in the case where there is no response code - because the error occurred while writing the request. Changing the log level from WARN to DEBUG is not sufficient to fix this IMO.
@craigmcnally, it's not correct. 
There re-request can be triggered by two reasons: 
1) when the srs returns a full response but with incorrect status code
2) when there a problem in srs with passing the body and the timeout exception is thrown (or any else potential http exceptions).
These two cases are different situations and they should have different log messages. I corrected it and we didn't pass the hardcoded "400" value. 
image

src/main/java/org/folio/oaipmh/client/SourceStorageSourceRecordsClient.java
IlliaDaliek: I added a test.

MikhailFokanov (approved):

craigmcnally:
> > @IlliaDaliek I noticed that you've introduce new changes here after the PR was approved. Please re-request reviews from those that previously approved. This feels a little like scope creep here. Strictly speaking this probably should have happened in a separate PR, probably with its own JIRA. Given that the changes are already here, we might as well go with it.
> 
> @craigmcnally, there no serious changes, I moved srs idle timeout to module settings as Mikhail asked me to do that. As well when I verified the "schema update testing" ticket I noticed incorrect logs and fixed them as well in the current PR. And among the code the laterals were used instead of already defined constants, I corrected this too.

OK thanks.  I can't stress this enough how important it is to communicate these details in the PR.  Otherwise, it's very hard to understand what's going on, and why.

craigmcnally (approved):

PR #196, opened 2021-03-12
MODOAIPMH-296: Update dependencies in mod-oai-pmh before Iris release

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-296

### PURPOSE
Update oai-pmh dependencies before Iris release. 

### CHANGES
**Dependencies updated:**
- vertx-completable-future removed
- mod-configuration-client 5.4.0 -> 5.6.0
- mod-source-record-storage-client.version 4.0.1 -> 4.1.4

The CompletableFuture replaced with vertx Future.

ModuleDescriptor will be updated at the release PR.

### DIFF
--- a/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
@@ -106,7 +106,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
             }
           });
         } else {
-          logger.error(request.getVerb().value() + " response from SRS status code: {}: {}", response.statusMessage(), response.statusCode());
+          logger.error(request.getVerb().value() + " response from SRS status code: {0}: {1}", response.statusMessage(), response.statusCode());
           throw new IllegalStateException(response.statusMessage());
         }
       } catch (Exception e) {
@@ -121,7 +121,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
     JsonArray instances = storageHelper.getItems(instancesResponseBody);
     Integer totalRecords = storageHelper.getTotalRecords(instancesResponseBody);
 
-    logger.debug("{} entries retrieved out of {}", instances != null ? instances.size() : 0, totalRecords);
+    logger.debug("{0} entries retrieved out of {1}", instances != null ? instances.size() : 0, totalRecords);
 
     if (request.isRestored() && !canResumeRequestSequence(request, totalRecords, instances)) {
       OAIPMH oaipmh = getResponseHelper().buildBaseOaipmhResponse(request).withErrors(new OAIPMHerrorType()
@@ -176,7 +176,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
             try {
               record.withMetadata(buildOaiMetadata(request, source));
             } catch (Exception e) {
-              logger.error("Error occurred while converting record to xml representation.", e, e.getMessage());
+              logger.error("Error occurred while converting record to xml representation. {0}", e, e.getMessage());
               logger.debug("Skipping problematic record due the conversion error. Source record id - " + recordId);
               return;
             }
@@ -223,7 +223,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
 
   protected void addRecordsToOaiResponse(OAIPMH oaipmh, Collection<RecordType> records) {
     if (!records.isEmpty()) {
-      logger.debug("{} records found for the request.", records.size());
+      logger.debug("{0} records found for the request.", records.size());
       oaipmh.withListRecords(new ListRecordsType().withRecords(records));
     } else {
       oaipmh.withErrors(createNoRecordsFoundError());
--- a/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
+++ b/src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
@@ -106,7 +106,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
             }
           });
         } else {
-          logger.error(request.getVerb().value() + " response from SRS status code: {0}: {1}", response.statusMessage(), response.statusCode());
+          logger.error(request.getVerb().value() + " response from SRS status code: {}: {}", response.statusMessage(), response.statusCode());
           throw new IllegalStateException(response.statusMessage());
         }
       } catch (Exception e) {
@@ -121,7 +121,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
     JsonArray instances = storageHelper.getItems(instancesResponseBody);
     Integer totalRecords = storageHelper.getTotalRecords(instancesResponseBody);
 
-    logger.debug("{0} entries retrieved out of {1}", instances != null ? instances.size() : 0, totalRecords);
+    logger.debug("{} entries retrieved out of {}", instances != null ? instances.size() : 0, totalRecords);
 
     if (request.isRestored() && !canResumeRequestSequence(request, totalRecords, instances)) {
       OAIPMH oaipmh = getResponseHelper().buildBaseOaipmhResponse(request).withErrors(new OAIPMHerrorType()
@@ -176,8 +176,8 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
             try {
               record.withMetadata(buildOaiMetadata(request, source));
             } catch (Exception e) {
-              logger.error("Error occurred while converting record to xml representation. {0}", e, e.getMessage());
-              logger.debug("Skipping problematic record due the conversion error. Source record id - " + recordId);
+              logger.error("Error occurred while converting record to xml representation. {}", e, e.getMessage());
+              logger.debug("Skipping problematic record due the conversion error. Source record id - {}.", recordId);
               return;
             }
           } else {
@@ -223,7 +223,7 @@ public abstract class AbstractGetRecordsHelper extends AbstractHelper {
 
   protected void addRecordsToOaiResponse(OAIPMH oaipmh, Collection<RecordType> records) {
     if (!records.isEmpty()) {
-      logger.debug("{0} records found for the request.", records.size());
+      logger.debug("{} records found for the request.", records.size());
       oaipmh.withListRecords(new ListRecordsType().withRecords(records));
     } else {
       oaipmh.withErrors(createNoRecordsFoundError());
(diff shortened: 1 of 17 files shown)


illia-borysenko (approved):

IlliaDaliek:
The coverage is 73% but this coverage relates to the new code. OaiPmhSetImpl has changes that contain a correction of log messages and I don't think that here any further unit tests are needed.

Igor-Gorchakov (approved):

craigmcnally:
Can you please explain why in some places were using log4j2 style lazy evaluation but in other places were using string.format %s for log messages?

craigmcnally:
It would be helpful to have a high level description of what you're doing here with logging changes

craigmcnally:
> Can you please explain why in some places were using log4j2 style lazy evaluation but in other places were using string.format %s for log messages?
> 
> Examples:
> https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationUtil.java#L67
> 
> And
> 
> https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/rest/impl/ModTenantAPI.java#L164
> 
> And here you're switching from one to the other, and as a slight variation this class has a static import of String.format
> https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java#L74
> 
> I'm _guessing_ that the lazy eval { } ones aren't doing what you think they are, and should instead use string.format until we upgrade to log4j2. In any event, we need to bring some consistency to this and ensure log messages are actually being properly formed. It's not very helpful to have a bunch of curly braces all over the place in the log ;)

Looking around s bit I see we are using log4j2, or at least we have log4j2.properties files.  If that's the case why wouldn't we use lazy evaluation instead of String.format?

This is confusing to me. I'll be quiet and give you a chance to explain what's going on here.

IlliaDaliek:
> > Can you please explain why in some places were using log4j2 style param substitution/lazy evaluation but in other places were using string.format %s for log messages?
> > Examples:
> > https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/oaipmh/helpers/RepositoryConfigurationUtil.java#L67
> > 
> > And
> > https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/rest/impl/ModTenantAPI.java#L164
> > 
> > And here you're switching from one to the other, and as a slight variation this class has a static import of String.format
> > https://github.com/folio-org/mod-oai-pmh/blob/ec594af1f2a9f186225f3ef8712eecd90bb60f22/src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java#L74
> > 
> > I'm _guessing_ that the lazy eval { } ones aren't doing what you think they are, and should instead use string.format until we upgrade to log4j2. In any event, we need to bring some consistency to this and ensure log messages are actually being properly formed. It's not very helpful to have a bunch of curly braces all over the place in the log ;)
> 
> Looking around s bit I see we are using log4j2, or at least we have log4j2.properties files. If that's the case why wouldn't we use the log4j2 param substitution/lazy evaluation instead of String.format?
> 
> This is confusing to me. I'll be quiet and give you a chance to explain what's going on here.


Hi @craigmcnally. I replaced the circle braces with "String.format(..)" because they weren't replaced with values passed to log method parameters. But I figured out that we should put the value starting from zero to circle braces to point a position for values to be put in the log message. I did such refactoring among the code solving this issue. As well in some places was an incorrect order of parameters for error logs, firstly we should pass a throwable object and then objects to be put in the log message. I tested these changes locally to ensure that the logs are correct.

craigmcnally:
> Hi @craigmcnally. I replaced the circle braces with "String.format(..)" because they weren't replaced with values passed to log method parameters. But I figured out that we should put the value starting from zero to circle braces to point a position for values to be put in the log message. I did such refactoring among the code solving this issue. As well in some places was an incorrect order of parameters for error logs, firstly we should pass a throwable object and then objects to be put in the log message. I tested these changes locally to ensure that the logs are correct.

OK, I think I see what's happening here... the `{N}` syntax was a hint.  By default Vertx uses JUL logging.  If you don't specify an override via `vertx.logger-delegate-factory-class-name` that's what will be used.  If you look at the pom file you can see that we specify `io.vertx.core.logging.Log4j2LogDelegateFactory`:  https://github.com/folio-org/mod-oai-pmh/blob/90ad79589b58c83c8c5e722c93e07a1123a560c1/pom.xml#L664

Is it the case where you're running your unit tests via an IDE which doesn't provide this system property, and therefore you get the JUL logging?

IlliaDaliek:
> > Hi @craigmcnally. I replaced the circle braces with "String.format(..)" because they weren't replaced with values passed to log method parameters. But I figured out that we should put the value starting from zero to circle braces to point a position for values to be put in the log message. I did such refactoring among the code solving this issue. As well in some places was an incorrect order of parameters for error logs, firstly we should pass a throwable object and then objects to be put in the log message. I tested these changes locally to ensure that the logs are correct.
> 
> OK, I think I see what's happening here... the `{N}` syntax was a hint. By default Vertx uses JUL logging. If you don't specify an override via `vertx.logger-delegate-factory-class-name` that's what will be used. If you look at the pom file you can see that we specify `io.vertx.core.logging.Log4j2LogDelegateFactory`:
> 
> https://github.com/folio-org/mod-oai-pmh/blob/90ad79589b58c83c8c5e722c93e07a1123a560c1/pom.xml#L664
> 
> Is it the case where you're running your unit tests via an IDE which doesn't provide this system property, and therefore you get the JUL logging?


@craigmcnally , I verified the logs by real calls to mod-oai-pmh. I used run IDE configuration and I didn't specify thevertx logger property.
image

craigmcnally:
> I should provide a Log4j2LogDelegateFactory in option and adjust the log messages to use it. Am I correct?

That's right

IlliaDaliek:
> > I should provide a Log4j2LogDelegateFactory in option and adjust the log messages to use it. Am I correct?
> 
> That's right


Hi @craigmcnally, I set up a run configuration to use Log4j2LogDelegateFactory and corrected the log messages accordingly. From those things that you already viewed I refactored a couple of lines to use circle braces instead of String.format in logs at ModTenantApi class, and added a method to transform the Java dto to json string in order to log a real JSON body instead of the object name in java memory (like org.folio.rest.jaxrs.model.FolioSet@2234).

src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
```diff
@@ -106,7 +106,7 @@ protected void requestAndProcessSrsRecords(Request request, Context ctx, Promise
             }
           });
         } else {
-          logger.error(request.getVerb().value() + " response from SRS status code: {0}: {1}", response.statusMessage(), response.statusCode());
+          logger.error(request.getVerb().value() + " response from SRS status code: {}: {}", response.statusMessage(), response.statusCode());
```
craigmcnally: Where this is an error it's unlikely to make a big difference, but generally its better to use lazy logging...

```
logger.error("{} response from SRS status code: {}: {}",
    () -> request.getVerb().value(),
    () -> response.statusMessage(),
    () -> response.statusCode());
```

src/main/java/org/folio/rest/impl/OaiPmhImpl.java
```diff
@@ -137,7 +137,7 @@ public void getOaiRecords(String verb, String identifier, String resumptionToken
             verbHelper
               .handle(request, vertxContext)
               .compose(response -> {
-                logger.debug(verb + " response: {0}", response.getEntity());
+                logger.debug(verb + " response: {}", response.getEntity());
```
craigmcnally:
```suggestion
                logger.debug("{} response: {}", verb,  () -> response.getEntity());
```

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
```diff
@@ -71,7 +73,7 @@ public void putOaiPmhSetsById(String id, String lang, FolioSet entity, Map<Strin
       Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     vertxContext.runOnContext(v -> {
       try {
-        logger.info("Put set by id with id: '{0}' and body: {1}", id, entity);
+        logger.info("Put set by id with id: '{}' and body: {}", id, entityToJsonString(entity));
```
craigmcnally, line 76:
```suggestion
        logger.info("Put set by id with id: '{}' and body: {}", id, () -> entityToJsonString(entity));
```

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
```diff
@@ -90,7 +92,7 @@ public void postOaiPmhSets(String lang, FolioSet entity, Map<String, String> oka
       Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     vertxContext.runOnContext(v -> {
       try {
-        logger.info("Post set with body: {0}", entity);
+        logger.info("Post set with body: {}", entityToJsonString(entity));
```
craigmcnally, line 95:
```suggestion
        logger.info("Post set with body: {}", () -> entityToJsonString(entity));
```

craigmcnally (approved):
A few suggestions,.  Approving now so we don't need another review iteration

src/main/java/org/folio/oaipmh/helpers/AbstractGetRecordsHelper.java
IlliaDaliek: Okay, thank you for the advice!

src/main/java/org/folio/rest/impl/OaiPmhSetImpl.java
IlliaDaliek: I tried the approach but the compiler swears on it.

I will back to it at the bugfix ticket PR's

craigmcnally:
I see you chose not to commit my suggestions, and I wanted to make sure you understand the implication of these changes as they may impact performance.  https://github.com/folio-org/mod-oai-pmh/pull/196#discussion_r595140129 for instance, that encoding JSON as a string will happen all the time, even if debug logging isn't enabled.  By calling the method via a lambda, it would only be executed if debug logging was enabled.  Where these responses are potentially massive, this could be a non-trivial operation.  Furthermore, it doesn't make sense to mix plain old string concatenation with parameter substitution.  

I had approved to avoid needing another review cycle, but was expecting the changes to be committed prior to you merging the PR.  I can see now that expectation wasn't clear.

PR #198, opened 2021-03-29
MODOAIPMH-303: Add index to mod-oai-pmh.instances table

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-303

### PURPOSE
Add index on id column which increases querying performance. 

As well in this PR I moved the scripts related to the instances table to the "create-instance-ids-table.xml" to keep all related changes in one file.

### DIFF
--- a/src/main/resources/liquibase/tenant/scripts/2020-07-16--10-00-create-instnace-ids-table.xml
+++ b/src/main/resources/liquibase/tenant/scripts/2020-07-16--10-00-create-instnace-ids-table.xml
@@ -59,6 +59,11 @@
   </changeSet>
 
   <changeSet id="2021-03-29--14-00-add-index-on-id-pk-field" author="Illia Daliek">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <indexExists tableName="instances" indexName="instance_id_pk_idx"/>
+      </not>
+    </preConditions>
     <createIndex tableName="instances" indexName="instance_id_pk_idx">
       <column name="id"/>
     </createIndex>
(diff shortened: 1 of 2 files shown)


src/main/resources/liquibase/tenant/scripts/2020-07-16--10-00-create-instnace-ids-table.xml
```diff
@@ -18,10 +18,50 @@
...
+  </changeSet>
+
   <changeSet id="2021-02-24--12-00-add-autoIncrement-pk" author="Illia Daliek">
     <addColumn tableName="instances">
       <column name="id" type="integer" autoIncrement="true"/>
     </addColumn>
   </changeSet>
 
+  <changeSet id="2021-03-29--14-00-add-index-on-id-pk-field" author="Illia Daliek">
+    <createIndex tableName="instances" indexName="instance_id_pk_idx">
```
craigmcnally, line 62: why not add the precondition here too?

TotoroDevEbsco: @craigmcnally, I moved some scripts from files and they were already applied at environments, but the scripts moving triggered their execution again which leads to an error. The instance_id_pk_idx is totally new here and has not been executed anywhere, that is why it is not required the precondition.

illia-borysenko (approved):

src/main/resources/liquibase/tenant/scripts/2020-07-16--10-00-create-instnace-ids-table.xml
craigmcnally: not required, but it can't hurt, In the case where maybe the module is enabled, then disabled, and later enabled again I think not having this here would be problematic.

IlliaDaliek: Okay, I added the precondition.

PR #199, opened 2021-03-30
MODOAIPMH-302: Response encoding does not use UTF-8 represenation of Unicode

TotoroDevEbsco:
Jira: https://issues.folio.org/browse/MODOAIPMH-302

### PURPOSE

Fix encoding to display the Cyrillic symbols in the correct manner.

### APPROACH

By default, the MracXmlWriter which translates records to XML uses the AnselToUnicode converter which cannot translate Cyrillic symbols and many others as well in a proper way. We should escape the conversion of each field/subfield data using such converter type. I tested it with both MARC8 and UTF-8 records and the output was good.

**Place where the subfield data is converted.**

encoding-1

**Here we see that data is correct and has a normal view before conversion.** 

before converter

**After the conversion, we see that the symbols cannot be translated and it was replaced with the Unicode sequence**

after converter

**In converter logic, we see that the case with Russian literals is unexpected here and it says that "Don't know what to do with this character.  No Unicode equivalent.  Encode it in the output as <U+XXXX>, where XXXX is the Marc character."**

result


illia-borysenko (approved):

src/test/resources/inventory_view/instance_ids_underlying_srs_records_with_cyrillic.json
```diff
@@ -0,0 +1 @@
+{  "instanceId": "ebbb759a-dd08-4bf8-b3c3-3d75b2190c41", "source": "FOLIO", "updatedDate": "2020-06-15T11:07:48.563Z",  "deleted": "false",  "suppressFromDiscovery": "false"}{  "instanceId": "ac9ac919-a35e-4a2d-b0a7-8cb5a63ea20d", "source": "FOLIO", "updatedDate": "2020-06-15T11:07:48.732Z",  "deleted": "false",  "suppressFromDiscovery": "false"}
```
craigmcnally, line 1: is there supposed to be a newline between the two records?

craigmcnally: or a comma?  both?

craigmcnally: never mind, I guess not.  The API docs aren't very clear about this, but I tried it and this looks right.

craigmcnally:
I don't see any problems with the code changes here, but it's not clear exactly what the problem was, or how it was solved.  I'm deducing that this has something to do with the version of mar4j we were using, and that by upgrading that dependency and providing those two additional arguments to the `MarcXmlWriter.writeSingleRecord(...)` call solves the problem.  Since there aren't any details in the JIRA, can you please provide some details in the PR description?

craigmcnally:
> I don't see any problems with the code changes here, but it's not clear exactly what the root cause was, or how it was solved. I'm deducing that this has something to do with the version of mar4j we were using, and that by upgrading that dependency and providing those two additional arguments to the `MarcXmlWriter.writeSingleRecord(...)` call solves the problem. Since there aren't any details in the JIRA, can you please provide some details in the PR description?

Thank you for adding context/explanation to the description @TotoroDevEbsco

PR #200, opened 2021-03-30
MODOAIPMH-293: Transfer suppressed records with discovery flag is not honored for instances suppressed from discovery

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-293

### PURPOSE

Fix misconfiguration issue caused by the reason that downloading instances is performed in a separate context which wasn't populated with oai-pmh configs data.


andrii-novytskyi (approved):

illia-borysenko (approved):

craigmcnally:
Why do we need a separate "downloadContext" (line 140)?

craigmcnally (changes requested):
Please add a unit test to verify this fix

craigmcnally:
I think this demonstrates that this won't work as-is:

```
package org.folio.rest.impl;
import static java.util.Objects.nonNull;
import io.vertx.core.Context;
import io.vertx.core.Vertx;
import io.vertx.core.WorkerExecutor;

public class ContextStuff {

  public static void main(String[] args) throws Exception {
    String tenant = "foo";

    Vertx vertx = Vertx.vertx();
    Context downloadContext = vertx.getOrCreateContext();
    System.out.println("downloadContext... " + downloadContext);

    WorkerExecutor executor = vertx.createSharedWorkerExecutor("saving-executor", 5);

    vertx.runOnContext(v -> {
      Context vertxContext = Vertx.currentContext();

      // e.g. loadConfiguration(... vertxContext)
      vertxContext.config().put(tenant, "zippity zap");

      System.out.println("vertxContext passed into handle(...) " + vertxContext);

      // later in MarcWithHoldingsRecordHandler(... vertxContext):
      executor.executeBlocking(p -> {
        Context current = Vertx.currentContext();
        System.out.println("in executeBlocking... " + current);

        setupCurrentContextConfigFromContext(vertxContext, tenant);
        download(downloadContext, tenant);
        p.complete();
      }, done -> {
        System.out.println("in executeBlocking resultHandler... " + Vertx.currentContext());
        vertx.close();
      });
    });
  }

  public static void download(Context ctx, String tenant) {
    System.out.println("in download... " + ctx);
    String config = ctx.config().getString(tenant);
    System.out.println("tenant config: " + config);
  }

  private static void setupCurrentContextConfigFromContext(Context fromContext, String tenant) {
    String config = fromContext.config().getString(tenant);
    if (nonNull(config)) {
      Vertx.currentContext().config().put(tenant, config);
    }
  }

}
```

When run:
```
downloadContext... io.vertx.core.impl.EventLoopContext@12d2ce03
vertxContext passed into handle(...) io.vertx.core.impl.EventLoopContext@6b15334e
in executeBlocking... io.vertx.core.impl.EventLoopContext@6b15334e
in download... io.vertx.core.impl.EventLoopContext@12d2ce03
tenant config: null
in executeBlocking resultHandler... io.vertx.core.impl.EventLoopContext@6b15334e
```

craigmcnally:
Honestly, I feel like storing the configuration on a vertx context just adds a layer of complexity that isn't necessary.  Really, all we need is a static map with tenant as the key and the config as the value.  There's no need to store this on any vertx context.  Doing so is only going to lead to these sorts of problems.  

We already have `public static String RepositoryConfigurationUtil.getProperty(tenant, name)`, why not just store the configs in a static map in RepositoryConfigurationUtil?  IMO there's no need to get the vert contexts involved here.

IlliaDaliek:
> Honestly, I feel like storing the configuration on a vertx context just adds a layer of complexity that isn't necessary. Really, all we need is a static map with tenant as the key and the config as the value. There's no need to store this on any vertx context. Doing so is only going to lead to these sorts of problems.
> 
> We already have `public static String RepositoryConfigurationUtil.getProperty(tenant, name)`, why not just store the configs in a static map in RepositoryConfigurationUtil? IMO there's no need to get the vert contexts involved here.

Hi @craigmcnally, I strongly agree with you that it will be much simpler and less problematic, I will refactor this part of the code.

IlliaDaliek:
> Please add a unit test to verify this fix

Unit test has been added

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -229,7 +229,7 @@ private void processBatch(Request request, Context context, Promise<Response> oa
           return;
         }
 
-        String nextInstanceId = instances.size() < batchSize ? null : instances.get(batchSize).getString(INSTANCE_ID_FIELD_NAME);
+        String nextInstanceId = instances.size() <= batchSize ? null : instances.get(batchSize).getString(INSTANCE_ID_FIELD_NAME);
```
IlliaDaliek, line 232: Here was a bug
When the total count of records in a system is equal to batch size the index out of bounds exception is thrown. 
If we have for example 10 records in total and batch size 10 as well, then the first condition will fail and we will try to get id of 11's instance which is absent in the list.

illia-borysenko (approved):

craigmcnally (approved):

PR #201, opened 2021-03-31
MODOAIPMH-305: <datestamp> element not populated correctly when harvesting with marc21_withholdings prefix

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-305

### PURPOSE
Fix the datestamp element to display the real record date instead of 1970-01-01

As well there was an issue in the datestamp values, it differs from the real srs record created/updated date. The records are stored in BD using the UTC date time and we have to display the same dates instead of UTC + harvester difference in time with UTC.


src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -544,7 +544,7 @@ private void enrichDiscoverySuppressed(JsonObject itemsandholdingsfields, JsonOb
       ).forEach(instance -> {
       final String instanceId = instance.getString(INSTANCE_ID_FIELD_NAME);
```
craigmcnally, line 545: I realize this isn't something introduced in this PR but can you please fix the indentation here?

craigmcnally:
I just looked it up in the spec and it should always be UTC... so this is a good call.

craigmcnally (approved):

craigmcnally:
Was this merged even though it didn't pass checks? 
image

IlliaDaliek:
> Was this merged even though it didn't pass checks?
> image

Hi @craigmcnally , because I formatted the method and the lines started to be considered as "changed" which affects total coverage in PR but it will not affect the general coverage.  Another code was covered already. No need to put tests here to cover "formatted lines".

PR #205, opened 2021-04-09
MODOAIPMH-312: Rerequesting of SRS doesn't work properly

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-312

### PURPOSE

Fix issues with an increasing number of attempts to re-request srs in two times after each exception handler.

### DIFF
--- a/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
+++ b/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
@@ -196,7 +197,7 @@ public class MarcWithHoldingsRequestHelper extends AbstractHelper {
     try {
       boolean deletedRecordSupport = RepositoryConfigurationUtil.isDeletedRecordsEnabled(request.getRequestId());
       int batchSize = Integer.parseInt(
-        RepositoryConfigurationUtil.getProperty(request.getTenant(),
+        RepositoryConfigurationUtil.getProperty(request.getRequestId(),
           REPOSITORY_MAX_RECORDS_PER_RESPONSE));
 
       getNextInstances(request, batchSize, context, requestId, firstBatch).future().onComplete(fut -> {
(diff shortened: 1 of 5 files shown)


prokhorovalexey (approved):
But let's discuss verbally first.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -738,7 +743,12 @@ private void doPostRequestToSrs(SourceStorageSourceRecordsClient srsClient, Vert
           return;
         }
         srsResponse.bodyHandler(buffer -> handleSrsResponse(srsClient, promise, buffer));
-      }, e->  retrySRSRequest(srsClient, vertx, deletedRecordSupport, listOfIds, attemptsCount, retryAttempts, promise, 0, null));
+      }, e-> {
+        if (!retryTriggered.get()) {
```
MikhailFokanov, line 747: Do it atimically, otherwise you have race condition

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -719,12 +720,16 @@ private HttpClientRequest createEnrichInventoryClientRequest(HttpClient httpClie
...
     try {
+      AtomicBoolean retryTriggered = new AtomicBoolean();
       srsClient.postSourceStorageSourceRecords("INSTANCE", deletedRecordSupport, listOfIds, srsResponse -> {
         int statusCode = srsResponse.statusCode();
         String statusMessage = srsResponse.statusMessage();
         srsResponse.exceptionHandler(e -> {
-          logger.error("Error has been occurred while requesting SRS: " + e.getMessage(), e);
-          retrySRSRequest(srsClient, vertx, deletedRecordSupport, listOfIds, attemptsCount, retryAttempts, promise, statusCode, statusMessage);
+          logger.error("Error has been occurred while requesting SRS kek1: " + e.getMessage(), e);
+          if (!retryTriggered.get()) {
```
MikhailFokanov, line 729: Do it atimically, otherwise you have race condition

MikhailFokanov (changes requested):

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: Thanks, done

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
IlliaDaliek: Thanks, done

MikhailFokanov (approved):

craigmcnally:
Why weren't unit tests updated for this?  It's important that we're exercising code paths which have been problematic/buggy.  Cleary our tests didn't catch this bug, so they need to be updated.  @TotoroDevEbsco Please create another PR with tests for this fix.

IlliaDaliek:
> Why weren't unit tests updated for this? It's important that we're exercising code paths which have been problematic/buggy. Cleary our tests didn't catch this bug, so they need to be updated. @TotoroDevEbsco Please create another PR with tests for this fix.

Yes, we will have another PR with adjusted tests but now we don't have time for that. That is the reason why they weren't updated.

craigmcnally:
I don't think it's acceptable to treat proper testing as an afterthought.  The PR never should have been merged w/o this.  The JIRA should never have been closed w/o this, and a release definitely should not have been made.

magdazacharska:
I agree with @craigmcnally - at this point we cannot afford inadvertently introducing bugs because of the lack of the proper testing.

PR #218, opened 2021-07-14
MODOAIPMH-337: The mod-oai-pmh schema is not populated correctly after RMB 32 update. 

IlliaDaliek:
Jira: https://issues.folio.org/browse/MODOAIPMH-337

### PURPOSE
Move the liquibase scripts to loadData method which will resolve the problem with the empty oai-pmh database.
Add tests
Improve exception handling

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,6 @@
     <aspectj.version>1.9.6</aspectj.version>
     <raml-module-builder.version>33.0.2</raml-module-builder.version>
     <vertx.version>4.1.0</vertx.version>
-    <vertx-folio.version>4.1.0-FOLIO</vertx-folio.version>
     <apache-httpclient.version>4.5.11</apache-httpclient.version>
     <apache-commons-lang3.version>3.9</apache-commons-lang3.version>
     <time-adapters.version>1.1.3</time-adapters.version>
--- a/pom.xml
+++ b/pom.xml
@@ -91,10 +91,24 @@
         <artifactId>vertx-pg-client</artifactId>
         <version>${vertx.version}-FOLIO</version>
       </dependency>
+      <dependency> <!-- needed for mockserver-client-java -->
+        <groupId>io.netty</groupId>
+        <artifactId>netty-bom</artifactId>
+        <version>4.1.65.Final</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
       <dependency>
         <groupId>org.junit</groupId>
         <artifactId>junit-bom</artifactId>
         <version>5.7.0-RC1</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>testcontainers-bom</artifactId>
+        <version>1.15.3</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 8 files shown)


src/test/java/org/folio/rest/impl/ModTenantAPIIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+
+@Disabled
+@Testcontainers
+class ModTenantAPIIntegrationTest {
```
julianladisch, line 31: The class name should be `ModTenantAPIIT` following the maven-surefire-plugin naming scheme:
https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html
This test uses the Dockerfile that requires mod-oai-pmh-fat.jar. The jar is not available when maven-failsave-plugin runs. It is available when maven-surefire-plugin runs.

pom.xml
```diff
@@ -131,21 +145,21 @@
     <aspectj.version>1.9.6</aspectj.version>
     <raml-module-builder.version>33.0.2</raml-module-builder.version>
     <vertx.version>4.1.0</vertx.version>
-    <vertx-folio.version>3.9.4-FOLIO</vertx-folio.version>
+    <vertx-folio.version>4.1.0-FOLIO</vertx-folio.version>
```
julianladisch, line 148: This is no longer in use and should be removed.

src/test/java/org/folio/rest/impl/ModTenantAPIUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+  @Test
+  void postTenantShouldSucceed(Vertx vertx, VertxTestContext vtc) {
+    modTenantAPI.postTenant(tenantAttributes, headers(), vtc.succeedingThenComplete(), vertx.getOrCreateContext());
+  }
+
+  @Disabled
+  @Test
+  void postTenantShouldFail_whenNoTenantId(Vertx vertx, VertxTestContext vtc) {
+    var headers = headers();
+    headers.remove("x-okapi-tenant");
```
julianladisch, line 101:
```suggestion
  @Test
  void postTenantShouldFail_whenNoOkapiUrl(Vertx vertx, VertxTestContext vtc) {
    var headers = headers();
    headers.remove("x-okapi-url");
```

src/test/java/org/folio/rest/impl/ModTenantAPIUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+  private TenantAttributes tenantAttributes = new TenantAttributes().withModuleTo("99.99.99");
+
+  @BeforeAll
+  void beforeAll(Vertx vertx, VertxTestContext vtc) {
+    var context = vertx.getOrCreateContext();
+    SpringContextUtil.init(vertx, context, ApplicationConfig.class);
+    vertx.runOnContext(v -> modTenantAPI = new ModTenantAPI());
+    PostgresClient.setPostgresTester(new PostgresTesterContainer());
+    PostgresClient.getInstance(vertx, OAI_TEST_TENANT).startPostgresTester();
+    startOkapiMockServer(vertx, vtc);
```
julianladisch, line 50: `new ModTenantAPI()` might be called after `beforeAll` has finished. To avoid this race condition the startOkapiMockServer need to go into the runOnContext.
```suggestion
    PostgresClient.setPostgresTester(new PostgresTesterContainer());
    PostgresClient.getInstance(vertx, OAI_TEST_TENANT).startPostgresTester();
    vertx.runOnContext(v -> {
      modTenantAPI = new ModTenantAPI());
      startOkapiMockServer(vertx)
      .onComplete(vtc.succeedingThenComplete());
    });
```

src/test/java/org/folio/rest/impl/ModTenantAPIUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+    vertx.close(res -> {
+      if(res.succeeded()) {
+        vtc.completeNow();
+      } else {
+        vtc.failNow(res.cause());
+      }
+    });
+  }
+
+  private Future<HttpServer> startOkapiMockServer(Vertx vertx, VertxTestContext testContext) {
```
julianladisch, line 64:
```suggestion
  private Future<HttpServer> startOkapiMockServer(Vertx vertx) {
```
This method returns a Future and therefore should not complete the testContext.
Better call like this:
```
startOkapiMockServer(vertx)
.onComplete(vtc.succeedingThenComplete());
```

src/test/java/org/folio/rest/impl/ModTenantAPIUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+    SpringContextUtil.init(vertx, context, ApplicationConfig.class);
+    vertx.runOnContext(v -> modTenantAPI = new ModTenantAPI());
+    PostgresClient.setPostgresTester(new PostgresTesterContainer());
+    PostgresClient.getInstance(vertx, OAI_TEST_TENANT).startPostgresTester();
+    startOkapiMockServer(vertx, vtc);
+  }
+
+  @AfterAll
+  void tearDown(Vertx vertx, VertxTestContext vtc) {
+    vertx.close(res -> {
```
julianladisch, line 55: vertx.close is not needed, VertxExtension automatically closes vertx.
If you really want to explicitly test that vertx.close succeeds use this:
```
vertx.close(vtc.succeedingThenComplete());
```
But I don't see a need for this, so the method tearDown should be deleted.

IlliaDaliek: Thank you for your suggestion.

src/test/java/org/folio/rest/impl/ModTenantAPIIntegrationTest.java
IlliaDaliek: Thank you, Julian.

pom.xml
IlliaDaliek: Good, I removed this property.

julianladisch (approved):
Looks good, thanks!

PR #220, opened 2021-09-03
MODOAIPMH-283 provide effective location and effective call number data when item record is not present

obozhko-folio:
Jira: https://issues.folio.org/browse/MODOAIPMH-283

### PURPOSE

The existing FOLIO implementation of the derived fields (effective location and effective call number) requires item records. The purpose of this story is to populate location and holdings information, following the logic of effective location/call number when the item data is not present.

### DIFF
--- a/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
+++ b/src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
@@ -451,7 +451,7 @@ public class MarcWithHoldingsRequestHelper extends AbstractHelper {
     httpRequest.putHeader(ACCEPT, APPLICATION_JSON);
     httpRequest.putHeader(CONTENT_TYPE, APPLICATION_JSON);
 
-    var enrichedInstancesStream = new BatchStreamWrapper(DATABASE_FETCHING_CHUNK_SIZE);
+    var itemsAndHoldingsStream = new BatchStreamWrapper(DATABASE_FETCHING_CHUNK_SIZE);
 
     //setup pgPool availability checker
     AtomicReference<SimpleConnectionPool<PooledConnection>> queue = new AtomicReference<>();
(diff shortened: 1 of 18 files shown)


alekGbuz (approved):

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -488,14 +506,20 @@ private void getNextBatch(String requestId, Request request, boolean firstBatch,
       try {
         for (JsonEvent jsonEvent : batch) {
           JsonObject value = jsonEvent.objectValue();
-          String instanceId = value.getString(ENRICHED_INSTANCE_ID);
-          Object itemsandholdingsfields = value.getValue(RecordMetadataManager.ITEMS_AND_HOLDINGS_FIELDS);
+          String instanceId = value.getString(INSTANCE_ID_FIELD_NAME);
+          Object itemsandholdingsfields = value;
```
khandramai, line 510: Why this variable is Object type if value-variable is JsonObject type?

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
```diff
@@ -488,14 +506,20 @@ private void getNextBatch(String requestId, Request request, boolean firstBatch,
...
           if (itemsandholdingsfields instanceof JsonObject) {
```
khandramai, line 511: itemsandholdingsfields equals to variable value. Why do we need to have two the same variables?

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
obozhko-folio: Yes, agree. Moreover, this variable is redundant.

src/main/java/org/folio/oaipmh/processors/MarcWithHoldingsRequestHelper.java
obozhko-folio: You are right, the redundant variable has been removed.

khandramai (approved):

PR #258, opened 2022-05-11
MODOAIPMH-417 - Handle DB timeouts on mod-oai-pmh

khandramai:
[MODOAIPMH-417](https://issues.folio.org/browse/MODOAIPMH-417) - Handle DB timeouts on mod-oai-pmh


khandramai:
Unit tests coverage is less then acceptable 80% due to technical restrictions. We need have DATABASE_FETCHING_CHUNK_SIZE configurable via system properties to cover all the if-branches of batches processing. This should be done in scope of  https://issues.folio.org/browse/MODOAIPMH-422. Current PR can be merged with current code coverage level as it will be improved.

alekGbuz (approved):

PR #262, opened 2022-05-24
MODOAIPMH-420 - Save UUIDs of failed instances

khandramai:
[PR#261](https://github.com/folio-org/mod-oai-pmh/pull/261) - Follow-up PR


khandramai:
Won't do

PR #264, opened 2022-05-30
MODOAIPMH-415 - Release 3.8.0

khandramai:
[MODOAIPMH-415](https://issues.folio.org/browse/MODOAIPMH-415) - Release 3.8.0

### DIFF
--- a/src/main/java/org/folio/oaipmh/dao/InstancesDao.java
+++ b/src/main/java/org/folio/oaipmh/dao/InstancesDao.java
@@ -5,6 +5,7 @@ import java.util.List;
 
 import org.folio.oaipmh.domain.StatisticsHolder;
 import org.folio.rest.jaxrs.model.RequestMetadataCollection;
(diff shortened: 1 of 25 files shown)


siarhei-charniak (approved):

alekGbuz (approved):

src/main/java/org/folio/oaipmh/dao/InstancesDao.java
```diff
@@ -17,6 +20,19 @@ public interface InstancesDao {
 
   Future<RequestMetadataLb> getRequestMetadataByRequestId(String requestId, String tenantId);
 
+  /**
+   * Returns request metadata collection
+   */
```
IlliaDaliek, line 25: Optional. 
Since other endpoints don't have a javadoc, because their purpose is clear from the name, this javadoc can be also removed

IlliaDaliek (approved):

PR #464, opened 2026-05-18
MODOAIPMH-600: Suppress from discovery flag in harvesting LINKED_DATA records

kjain110:


src/main/java/org/folio/oaipmh/helpers/records/RecordMetadataManager.java
```diff
@@ -102,9 +102,7 @@ private RecordMetadataManager() {
     electronicAccessPredicate = jsonObject -> {
       if (jsonObject.containsKey(ELECTRONIC_ACCESS_FILED_TAG_NUMBER)) {
         JsonObject dataFieldContent = jsonObject.getJsonObject(ELECTRONIC_ACCESS_FILED_TAG_NUMBER);
-        String firstIndicator = dataFieldContent.getString(FIRST_INDICATOR);
-        String secondIndicator = dataFieldContent.getString(SECOND_INDICATOR);
-        return StringUtils.isNotEmpty(firstIndicator) && StringUtils.isNotEmpty(secondIndicator);
+        return nonNull(dataFieldContent);
```
obozhko-folio, line 105: Please explain what is the reason to remove checking indicators here?

kjain110: We changed it so any existing 856 field gets t suppression flag even when indicators are blank/missing, preventing valid records from being skipped.

siarhei-charniak (approved):

obozhko-folio (approved):

